Issue #608, opened 2015-02-04
PyChess sets (pin, gin, availinfo) FICS vars to values that don't work well with other clients or telnet


pychess:
Original [issue 608](https://code.google.com/p/pychess/issues/detail?id=608) reported by [dbodom](https://code.google.com/u/112986820331174496732/) 2010-10-30

**What steps will reproduce the problem?**
1. Create a FICS account and play with your favorite telnet client
2. Download and start pychess, login to FICS using pychess with your previously created FICS account
3. Login again into your FICS account using telnet: pychess scrambled all your settings without asking your permission.

Settings changed includes also formula.

**What is the expected output? What do you see instead?**

I was expecting pychess to ASK me the permission to change my FICS settings, or even better, to adapt to them without making any change.

**Does it happen every time?**
Yes.

**What version of the product are you using?**
Lastest mercurial.

**Did you use an installed version of PyChess or did you run in from a**
**tarball/svn?**
I run it from source.

pychess:
Original [comment 1.](https://code.google.com/p/pychess/issues/detail?id=608#c1) posted by [mattgatto](https://code.google.com/u/mattgatto/) 2010-10-30

Thanks for the bug report. It's not practical for pychess to "ask you permission" to change each and every setting pychess needs for it to function correctly (there's alot of them), as pychess' goal is to be usable by people who don't care about var settings, in addition to people that do, like us. But it is supposed to save your var and formula settings settings (upon login) and restore them upon logout (ListAndVarManager.py), right Thomas?

I just tested tip, and gin and availinfo are the two most noticeable FICS var's that are not restored. I'm not sure about formulas since I haven't used a custom one for a long time.
- changed status to: Accepted
- added label: Defect

pychess:
Original [comment 2.](https://code.google.com/p/pychess/issues/detail?id=608#c2) posted by [dbodom](https://code.google.com/u/112986820331174496732/) 2010-10-30

Thank you Matt for your explanation on how pychess works.

I can see the bug on this workflow: it assumes that there will be a logout soon or later to restore the variables. But that's not true: what if pychess crashes or link gets broken?
The variables will not get restored, and, I guess, that was my case :)

General speaking, IMHO you can't assume that after a login there will certainly be a logout.
A possible workaround would be to save the vars into a static location (i.e. a file) _only_ the first time user logs in to FICS, then always restore variables from that file: even in case of a crash, the restored variables will be correct and a login/logout will fix the "telnet" environment.

[...]

PR #942, opened 2015-02-14
NoEscape turned off by default

Nazine:
It is troublesome for a user to lose by disconnection during a game on FICS. Setting noescape to off prevents this.


gbtami:
PyChess set only interface variables. We will never set any user variables in the future. We did it before, and it caused big trouble. See issue #608 They belong to the user and he can set it in the console.

Issue #256, opened 2015-02-03
Auto-save games options


pychess:
Original [issue 256](https://code.google.com/p/pychess/issues/detail?id=256) reported by [armand.delcros](https://code.google.com/u/106971933427282282316/) 2007-11-11

It will be nice to have options preferences for saving game.
For example, under xboard, I configure a destination directory for the
auto-save pgn game.

**What version of the product are you using?**

0.8.1
Did you use an installed version of PyChess or did you run in from a
tarball/svn?
tarball

pychess:
Original [comment 1.](https://code.google.com/p/pychess/issues/detail?id=256#c1) posted by [lobais](https://code.google.com/u/lobais/) 2007-11-11

So you say you'd like PyChess to like autosave once a minute?
Or you you mean like in glchess where games are automatically saved when you close
the client, and then reopened next time you start?
- removed label: Defect
- removed label: Medium
- added label: Enhancement
- added label: Low

pychess:
Original [comment 2.](https://code.google.com/p/pychess/issues/detail?id=256#c2) posted by [armand.delcros](https://code.google.com/u/106971933427282282316/) 2007-11-12

When my game is over, I would like that this game is automatically save as a pgn
file. (or a popup proposition). So it is not like glchess.

Thanks you !

pychess:
Original [comment 3.](https://code.google.com/p/pychess/issues/detail?id=256#c3) posted by [lobais](https://code.google.com/u/lobais/) 2007-11-12

So it is more like PyChess should manage a database of all games you've played?

pychess:
Original [comment 4.](https://code.google.com/p/pychess/issues/detail?id=256#c4) posted by [armand.delcros](https://code.google.com/u/106971933427282282316/) 2007-11-13

Why not, but sufficient for me is to save each game in a pgn file. It is not
necessary to have a full database concept.
I use the pgn file for a crafty's analyze after my game.

for example, in the preferences, i could choose the format of the filename for
example (for fics): 
"nickname-white-color"-"nickname-black-color".pgn 
and i could choose an auto save game option.

pychess:
Original [comment 5.](https://code.google.com/p/pychess/issues/detail?id=256#c5) posted by [lobais](https://code.google.com/u/lobais/) 2007-11-13

Ah I see.
What do you think about a preferences section, like the one attached?

If we were to implement a database and/or analyzing directly into PyChess, could you
describe or make a mockup of an ideal structure?
[...]

PR #957, opened 2015-04-06
new command line argument

gideon425:
The changes implement a new command line argument --auto_save_pgn_file

### DIFF
--- a/lib/pychess/Main.py
+++ b/lib/pychess/Main.py
@@ -466,7 +466,7 @@ class PyChess:
             glock.glock_connect_after(discoverer, "all_engines_discovered", do)
 
 def run (no_debug, no_glock_debug, no_thread_debug, log_viewer, chess_file,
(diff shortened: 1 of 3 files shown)


mattgatto:
Why not just save games automatically? I don't see the point of a command
option for this.

mattgatto:
We should consider just getting rid of the save dialog and just do this
automatically unless someone can think of a reason to do otherwise, in my
opinion.

gbtami:
I'm fine with getting rid of save dialog. Telling the truth I hate to close it every time I'm pychessing. But if we do that, we have to add some preferences option Thomas mentioned in the original issue #256. At least until we integrate the database stuff into the new GUI redesign.

gideon425:
I am open to modifying my changes.  I suggest we continue the discussion in issue 256. I will update that issue with my thinking

PR #976, opened 2015-06-12
Adding engines that contains spaces in file path

n-p:
tested at linux mint 17.1.


gbtami:
Fixed this by removing unnecessary shlex.split() in command.py instead.
Thx.  anyhow!

PR #1021, opened 2015-09-27
remove lowercasing of engine name to find pychess.py

### LINKED ISSUE #1020: bug : selecting pychess default engine causes traceback
  File "/media/data/root/Documents/chess/pychess/pychess/lib/pychess/widgets/enginesDialog.py", line 325, in selection_changed
    engine = discoverer.getEngineByName(name)
  File "/media/data/root/Documents/chess/pychess/pychess/lib/pychess/Players/engineNest.py", line 355, in getEngineByName
    return self.getEngines()[names.index(name)]
ValueError: 'Pychess.py' is not in list

bboutkov:
fixes #1020 

I tested it and it seems to work. The downside to this quick fix is that now the alphabetizing of engines is a little weird and the engines starting with a capital letter (seems to be only ours) appear at the top of the engine selection list. I dont know if this is a major concern.

### DIFF
--- a/lib/pychess/Players/engineNest.py
+++ b/lib/pychess/Players/engineNest.py
@@ -345,7 +345,7 @@ class EngineDiscoverer (GObject.GObject):
     
     def getEngines (self):
         """ Returns list of engine dicts """     
-        return sorted(self.engines, key=lambda engine: engine["name"].lower())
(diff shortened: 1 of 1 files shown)


bboutkov:
this issue is certainly much deeper than this proposed fix. closing.

Issue #1057, opened 2015-11-26
Visual reflection of session changes with respect to observed games and chat channels


cajone:
I watch a lot of observed games at the same time following relay events / teamleague games etc
If I'm concentrating  on one game I have to physically open another observed game to see if any comments have been whispered or if the board position has changed.

I think it would be a good enhancement if the game tag changed colour to reflect an update to that board while another board has focus. I was thinking with respect to observed boards that do not have focus to change the tag font colour to say light blue ( colour does not really matter) to reflect a whisper added and the font to change to red ( for non focused boards) if a move has been played.

To extend this idea to chat channels the same thing. It is only possible to view one channel at at time so again it would be nice for a visual reflection to happen to channels not in focus but have had content added to them.

For both the above cases once a channel of game has been gained focus then the colour drops back to it default value.

What do you think ?

( possible enhancement)  

P.S. I'm getting ahead of myself, I mentioned whispers above. I was thinking that the chat panel of a game notebook could be used to capture whispers intended for that game, currently this chat panel is not active if you are not a player of the game.

gbtami:
I like all.

cajone:
Ok I have come to a road block and currently cannot proceed ( need help ). I have added a push to my clone

https://github.com/cajone/pychess/tree/%231057

that represents the code I think should change the background colour of channels that have been updated with contents but are currently not in focus. THIS code segment is by far not the complete solution for this issue, but if I cannot get passed this bit I cant proceed.  

any help will be welcome

cajone:
Although channel highlighting is working on change of content, player/friends highlighting is still missing when content changes

PR #1091, opened 2015-12-20
#1057 changed func_data to a dictionary

cajone:
So as discussed I changed out the single func_data arg from a single string to a dictionary of
string:bool key,value pairs.

From the little testing I can do at this time ( not much chat traffic on fics at the moment ) it looks like most of the issues have gone.

The one I can still see is after to select a channel it turns to default foreground colour (as expected) but if you then hover over other highlighted channels they are also set to default.

I really can't think of a solution for this. If you can come up with something please implement, don't think I can look at this bit of code again in the coming days/weeks or months ( way to many hours for what should have been a relatively trivial problem )


cajone:
Changed the logic a bit and it seems to be functioning a little better will need monitored to clarify ( not a lot of test data flowing through fics last night)

gbtami:
All seems OK for me.

PR #1140, opened 2016-01-27
#docs started pylint cleanup 0.45->1.17/10

cajone:
BoardView.py is a large module containing most of the move logic for the board.

In order to be able to understand the code and be able to modify later 
I need to refactor / cleanup the module this will be on going as I understand a little more each time


ggjj1122:
yes yes yes do what needs to be done....i read between the lines my deppest
appologies :-)

PR #1142, opened 2016-01-27
#docs pylint cleanup 3.92->4.27/10

cajone:


gbtami:
Shouldn't be better to use
if (
instead of 
if(
I think the former is more readable.

cajone:
yes I will fix this, it must have got caught on the pattern match it was not intentional

gbtami:
OK, np.

PR #1358, opened 2016-05-23
Correct typographical error (sacrifies → sacrifices)

zuzak:
This commit changes instances of the word "sacrifies" to the more common word "sacrifices".
The former isn't in most dictionaries, but apparently used to be a word in Middle English.

Updated the message ID too.


gbtami:
Thx!

Issue #1026, opened 2015-10-14
feature request : secondary sort on games list


bboutkov:
Currently if I sort by game type, the games list displays games alphabetically by type and that is all. I think it would be useful to automatically also sort by rating, as sorting by game type alone isn't very useful for selecting a good game to observe.

Perhaps eventually we could have a subsort system where we sort based on column a, and then subsort that based on column b where a user selects the two columns a & b. These might be big dreams for now though ...

gbtami:
Changing our pixCompareFunction()  function in lib/pychess/ic/ICLounge.py to subsort on rating seems doable.
(If you like to experiment with it, it starts at line 599.)
https://python-gtk-3-tutorial.readthedocs.org/en/latest/treeview.html#setting-a-custom-sort-function

On the other side, Gtk.TreeView supports sorting by clicking on single columns only, so I can't imagine how can we implement two column selection at all.
https://python-gtk-3-tutorial.readthedocs.org/en/latest/treeview.html#sorting-by-clicking-on-columns

gbtami:
Implemented a multi-column sort based on TreeModelSortUtil class from
https://github.com/metomi/rose/blob/master/lib/python/rose/gtk/util.py
It saves previous sort column and reuses it as secondary sort column on next column selection.
Try it please!

cajone:
I believe this request has resolved I can use the column headers to sort on rating for white and then sort on rating for black and get back a list that shows the highest black rating versus white rating I can also sort by rating and then by type and this also seems to work

Issue #1126, opened 2016-01-09
chat panel lists jumping on long list selection


cajone:
If the group has a long list of users in it ie 'Other Players' oe 'Guests' and you try to select a user the list jumps, I would guess this is a the list centralising but I might be wrong

cajone:
Hi gbtami

With respect to our conversation over the weekend about simplifying the channel/player/Guest etc lists
into a single storelist. I have knocked up a very simplistic standalone panel

https://gist.github.com/cajone/5db621e7598d961935e1

my problem ( I just don't know how to do) would be to collapse both columns into 1 column. I got the impression you had an idea how to so this. 

If this is a completely different idea than the one you are thinking about lets chat on #pychess

gbtami:
I was thinking about using a custom cell renderer, something similar we use in sidepanel/bookPanel.py. The opening-engine-endgame parts share the same store of a treeview and a CellRendererPixbufXt is used to expand/collapse analysis rows.
Maybe this is overcomplication, and we should just use one treview as is.

cajone:
was trying to get to the heart of this problem. ( some thoughts)

What I observe is I can scroll the window of listviews till I can see the item I want to select, at this point everything seems to work as expected.

It's only at the point that I try to select the item that the jumping effect happens. My 'current' thinking is on selection of an item, an iterator is running so as to highlight the selection and as this iterator is traversing the list it is changing/refocusing the list with respect to the viewport. If we had a chat window that was vertically the same size as all the lists there would be no jumping, so the problem lies with selection and viewport

PR #1371, opened 2016-07-17
partial fix to #1026

cajone:
Still jumps of first selection, but seems to work as expected after that


cajone:
the above reference is wrong the correct issue is #1126

PR #1374, opened 2016-07-25
add new follow option on notify

bboutkov:
I think it can be useful to follow a friends games when they sign on, so I created a basic functionality for it, it seems to work fine on my local copy.

Hopefully I didn't break anything,  feel free to yell at me if I did something horrible!


gbtami:
Good idea. Thx Boris!

PR #1446, opened 2017-08-20
update

ahgm:


gbtami:
@ahgm pychess translations are maintained on https://www.transifex.com/projects/p/pychess/
Update your language translations there, please!

PR #1469, opened 2017-10-10
Minimal limit for MultiPV in panel Hints

ecrucru:
Hello,

In the panel Hints, you can force (without using +/-) a null or negative value for MultiPV. This check ensures that it cannot happen.

Also, if you set a zero, you cannot restore the value for the current game. That's a concern with the UI, or maybe a choice by design.

Regards.


gbtami:
Thx!

Issue #1454, opened 2017-09-16
Change the country from the UI


ecrucru:
Hello,

The country is part of `engines.json` but it has no corresponding changeable field in engineDialog.

For user-friendliness, is it possible to add that manual field ?

Thanks

ecrucru:
Hello,

I made an attempt to add the change of the country from the UI :

![image](https://user-images.githubusercontent.com/24614488/31576927-d0da0262-b105-11e7-8e8e-48ba8ab7a779.png)

That proposal has one issue : the combobox of the country doesn't reflect the current flag of the configuration when the selection is changed.

I need your help to finalize that feature.

Thanks.

gbtami:
Merged your updated PR. Thx!
It would be nice to have selection with keyboard from the combobox also. Unfortunately this is one of the oldest Gtk+ feature request, not addressed to date :( See https://bugzilla.gnome.org/show_bug.cgi?id=567141

PR #1475, opened 2017-10-14
Change the country from the UI

ecrucru:
In reference to #1454.


gbtami:
Cool! Some minor things:

You put `isoCountries.py` into `lib/pychess/System` but wants to import from `lib/pychess/Utils` in `enginesDialog.py`

The whole new initialization code of `engine_country_combo` in `def run(widgets):` should be moved to **before** `EnginesDialog(widgets)` line to not change back the very first engine country to Unknown.

In `def selection_changed(treeselection):` we need a line
`self.widgets["engine_country_combo"].set_active(0)`
before `for iso in ISO3166_LIST:` to show no flag when selecting engines with Unknown country.

In `def country_changed(widget):` after `discoverer.save()` you can update the flag in engines list with something like this:

```
                    path = addDataPrefix("flags/%s.png" % new_country)
                    if not os.path.isfile(path):
                        path = addDataPrefix("flags/unknown.png")

                    selection = self.tv.get_selection()
                    result = selection.get_selected()
                    if result is not None:
                        model, ts_iter = result
                        model[ts_iter][0] = get_pixbuf(path)
```

ecrucru:
Thank you for your detailed answer. The UI reacts correctly and the PR is updated.

Regards

PR #1485, opened 2017-10-22
Don't load the incorrect engines

ecrucru:
Hello,

When you add an incorrect engine, the error popup is not shown. It also generates an incorrect entry in `engines.json` which cannot be removed from the UI :

```
,
 {
  "command": "/home/komodo-9.02-linux",
  "country": "unknown",
  "md5": "ca1d3b4018540ee0c9a3c66cae5c1229",
  "name": "komodo-9.02-linux",
  "protocol": "xboard",
  "recheck": true
 }
```

It looks like a problem in the indent of the existing code. Here the fix which provides the expected output to the player.

Regards

### DIFF
--- a/lib/pychess/widgets/enginesDialog.py
+++ b/lib/pychess/widgets/enginesDialog.py
@@ -284,26 +284,27 @@ class EnginesDialog():
                                 break
                             else:
                                 continue
-                            if not check_ok:
-                                # restore the original
-                                engine = discoverer.getEngineByName(
-                                    self.cur_engine)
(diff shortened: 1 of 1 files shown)


gbtami:
Thx!

gbtami:
@ecrucru may I ask your real name? I like to add you to AUTHORS file. If you don't mind, of course.

ecrucru:
You may simply use the pseudo, because this is how I use to release the code. Thanks

gbtami:
Done.

PR #1486, opened 2017-10-22
Select the country with key event

ecrucru:
Following #1454, this alternative tends to select the first country matching with the pressed key.

The selection is indeed quicker.


ecrucru:
Also, is it possible to regenerate Transifex to start the translation of the 250 countries ?

Thanks.

gbtami:
Thx!

gbtami:
Updated translations template also. This will need some hours to be updated on transifex too.

PR #1487, opened 2017-10-25
Explicit NAG in the popup menu

ecrucru:
Hello,

This commit enhances the meaning of the NAG symbols in the popup menu of the annotation panel. The caption is translatable and the symbols are fetched from the internal dictionary without redundancy.

Regards


gbtami:
Cool, thx!

PR #1488, opened 2017-10-27
No selection of the unknown country

ecrucru:
Little correction for 9c72933dfeae00ba95e4b40ed92520921a1c108e.
"Unknown" should not be fetched if I press "U" in English (Uganda) or "I" in French (Iceland).
Unknown is voluntarily the first unsorted item of the list of the countries, so we don't loop on it.


gbtami:
Thx!

PR #1491, opened 2017-10-28
Refresh engineDialog with 1 engine only

ecrucru:
There is no change event in engineDialog if `engines.json` has 1 entry only.
The window appears as if there was no setting for the selected engine.
It affects both Windows and Linux.


gbtami:
Thx!

PR #1494, opened 2017-11-01
Exchange the players

ecrucru:
Hello,

This image explains the objective of that change :

image

Regards

### DIFF
--- a/glade/newInOut.glade
+++ b/glade/newInOut.glade
@@ -234,8 +234,8 @@
                                 <property name="valign">start</property>
                                 <property name="n_rows">4</property>
                                 <property name="n_columns">2</property>
-                                <property name="column_spacing">12</property>
-                                <property name="row_spacing">1</property>
+                                <property name="column_spacing">2</property>
+                                <property name="row_spacing">2</property>
                                 <child>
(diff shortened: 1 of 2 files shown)


gbtami:
I feel myself a little bit uncomfortable to add this Exchange button to New Game dialog. It seems cluttered with it for me. But.
After human-human and human-engine game end pychess offers "Play Rematch" button. Wouldn't it be enough to enable it on engine-engine game ands also?

ecrucru:
The button *Rematch* comes too late in time if you want to start the game with specific options (white/black, time, variant...). The dialog is rarely in the correct state, so manually switching the players makes up to 8 clicks of very poor interest because you have to exchange the player, select the engine in a possibly-long list and set the levels.

I agree that the button doesn't really help to have a concise layout. In your opinion, could an alternative be something like that ? Don't pay attention to the names of the engines, it is just a fake picture. A click on one of the new buttons exchanges the players.

image

You can indeed add the rematch button for the engines. I hadn't noticed that, good idea !

Thanks for your feedback.

gbtami:
Wow! This is much better. Go for it!

ecrucru:
The PR is updated : you should now be able to see the result.

Regards

gbtami:
Cool, thx!

PR #1495, opened 2017-11-05
No override of the deeper scores

ecrucru:
Hello,

It doesn't make sense to save the score without checking if the existing one has been computed deeper.
If you analyzed your game, now if you browse your game, you will not lose what was calculated.
It gives then a more consistent graph of the scores.

I also noticed that changing the position in the annotation panel should reset the current analysis in memory in order that it doesn't get copied from the previous move. May I ask you to verify that point ?

Regards


gbtami:
The fix is reasonable. Thx!
I don't understand the second part of your comment. Can you explain it a bit detailed please?

PR #1496, opened 2017-11-05
No override of the deep scores v2

ecrucru:
Hello,

Initially, this PR aims to add a popup menu item to reset all the existing evaluations, because if you relaunch the analysis with another analyzer, it will not necessarily override the existing evaluations. It looks OK.

However, the PR also fixes a regression I didn't noticed : after #1495, the analysis is stucked at depth 9 because pyChess returns the parsed depth as... a string (not an integer). So the added comparison was not fully effective. The problem mentioned in the previous PR is now fully reproducible :
- Play several moves
- Go back to the first move, and leave some time for the analysis to run as deep as possible
- Go to the next move (and so on) : it receives the residual evaluation of the previous position because this is based on the events and no link is made with the original node

I don't know how to fix that. That problem was not so obvious with the first version of the PR, that's why I am a bit disappointed at that situation (incomplete code).

Regards

### DIFF
--- a/lib/pychess/Utils/GameModel.py
+++ b/lib/pychess/Utils/GameModel.py
@@ -284,11 +284,11 @@ class GameModel(GObject.GObject):
             ply = analyzer.board.ply
             if score is not None:
                 if analyzer.mode == ANALYZING:
-                    if (ply not in self.scores) or (self.scores[ply][2] <= depth):
+                    if (ply not in self.scores) or (int(self.scores[ply][2]) <= int(depth)):
                         self.scores[ply] = (pv, score, depth)
                         self.emit("analysis_changed", ply)
                 else:
-                    if (ply not in self.spy_scores) or (self.spy_scores[ply][2] <= depth):
+                    if (ply not in self.spy_scores) or (int(self.spy_scores[ply][2]) <= int(depth)):
                         self.spy_scores[ply] = (pv, score, depth)
 
     def setOpening(self, ply=None):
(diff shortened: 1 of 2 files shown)


gbtami:
Maybe extending analysis tuples with engines board.ply can help here.
If you search for emit("analyze" in UCIEngine.py and in CECPEngine.py you can see how these lists of tuples are constructed now.

gbtami:
Fixed the "go to next move" problem in https://github.com/pychess/pychess/commit/efb8ac61f245f0deb2b3536129b7c8b939f104b4
Maybe it's a little bit hackish, but seems to work.

ecrucru:
Thanks for your proposed change. I just retested it and my impression with my slow VM is that the fix is not totally correct (at least for the UCI part). I try an explanation because on fast computers it may be not very visible.

An analysis starts with `go movetime xxx` or `go infinite`. The position and the ply should be locked at that moment until the engine dies or we receive a `bestmove`. When we change the position in the annotation panel, we immediately send a `stop` to end the infinite analysis (and eventually the time-limited search). The locked position/ply continues to be parsed whatever the number of info lines that the engine is finishing to produce (we can't assume it is equal to 1 message). The `bestmove` will unlock the current analyzed position and we start a new analysis with the unique pending ply (if any) which corresponds to another chess position. The analysis continues to be broadcasted with the ply added to the analysis tuple.

So you have added the ply in the tuple. It is a very good start to link the result with the right move, but we have to wait for the bestmove else the issue will still occur. I see no other reason. Is it hard to implement ?

Regards

gbtami:
Yes, you'r explanation is correct. My "fix" is not bullet proof and needs more work. Anyhow thx for testing it!

gbtami:
@ecrucru committed a better fix for UCI. Unfortunately CECP has no "bestmove" or similar so I just added a little sleep(0.1) after "exit" to stop analyze. Can you test it please?

ecrucru:
The UI reacts nicely for UCI and certainly for pychess-engine as well.
Thank you again for your excellent support !

PR #1506, opened 2017-11-18
Elo rating change in UI and PGN

ecrucru:
Hello,

Now that we can set the ELO in the properties of the game, we are able to get some related information about its evolution based on the time control of the game, the levels of the players, the chances to win, the termination... The properties of the game are then extended by the PR, as well as the export to PGN along with additional checks to secure the data (decoration for backslashes and quotes).

The concept is explained in the FIDE handbook and testable online at :
ratings.fide.com/calculator_rtd.phtml

The results are not the same than FICS.

Regards

### DIFF
--- a/glade/PyChess.glade
+++ b/glade/PyChess.glade
@@ -1055,10 +1055,13 @@
                           <object class="GtkTable" id="table1">
                             <property name="visible">True</property>
                             <property name="can_focus">False</property>
-                            <property name="n_rows">5</property>
+                            <property name="n_rows">6</property>
                             <property name="n_columns">4</property>
                             <property name="column_spacing">6</property>
                             <property name="row_spacing">4</property>
+                            <child>
+                              <placeholder/>
+                            </child>
                             <child>
                               <object class="GtkEntry" id="event_entry">
                                 <property name="visible">True</property>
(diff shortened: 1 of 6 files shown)


gbtami:
Nice work! Some minor thing I like to ask before merge.

Instead of "Statistics:" I suggest to use the more explicit "Rating change:" label, and changing the new `elo_stats` to `w_elo_change` and  `b_elo_change` aligned to the other entries of appropriate players. It would be nice to see only one number for white and one for black if gamemodel.status in (DRAW, WHITEWON, BLACKWON). If you like you can color them with green (for +) and red (for -).

Another thing I wanted to ask before. We like to follow Python PEP 8 style guidelines. To help ourself we use:
```
pip install flake8
pip install pep8-naming
flake8 --install-hook=git
git config flake8.strict true
```
This prevents committing non PEP 8 compliant changes.
We have some exceptions as you can see in https://github.com/pychess/pychess/blob/master/setup.cfg

You can configure several IDE to use these linters. I myself configured Geany following this: https://wiki.geany.org/howtos/check_python_code#using_flake8

ecrucru:
Please find the new version which includes your remarks.

For the validation of the code, I used the command line and read 300 messages. There is normally no message concerning the parts I changed, so tell me if additional changes are needed according to you.

Thanks

gbtami:
Thx!

PR #1508, opened 2017-11-20
Rotate the board in setup position

ecrucru:
Hello,

This is the first step to support the rotation of the board during the setup of a position. This is especially useful when you play as Black.

The next steps could be :
- the coordinates of the board in the left&bottom borders don't have yet a rotated equivalent on their opposite sides when the board is rotated
- once we start from the position, the principal board takes the orientation of the setup

Regards

### DIFF
--- a/glade/newInOut.glade
+++ b/glade/newInOut.glade
@@ -1493,21 +1493,6 @@
                                           </packing>
                                         </child>
                                         <child>
-                                          <object class="GtkToggleButton" id="side_button">
-                                            <property name="visible">True</property>
-                                            <property name="can_focus">True</property>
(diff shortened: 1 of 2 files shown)


gbtami:
Thx!

gbtami:
Telling the truth I was fighting this before but wasn't able to find a good solution. What I see you can use "Rotate board" menu item from main window "View" menu. But Ctrl+R is not working if the focus is on the dialog window.

Until we can find a solution to use Ctrl+R from global menu I think mowing the "Rotate board" button **above** to the board would be better as it changes only the view of the physical board not the game position as the group of widgets below the board.

PR #1509, opened 2017-11-21
Remove a comment quicker

ecrucru:
Hello,

This change aims to make the removal of a comment quicker in the annotation panel. The blank comments will also be considered as removed to eliminate the ambiguous caption of the popup menu "Edit comment" when the associated comment is blank.

Regards

### DIFF
--- a/lib/pychess/perspectives/games/annotationPanel.py
+++ b/lib/pychess/perspectives/games/annotationPanel.py
@@ -371,10 +371,22 @@ class Sidepanel:
         self.gamewidget.copy_pgn()
 
     def edit_comment(self, widget=None, board=None, index=0):
-        dialog = Gtk.Dialog(
-            _("Edit comment"), None, Gtk.DialogFlags.MODAL | Gtk.DialogFlags.DESTROY_WITH_PARENT, (
-                Gtk.STOCK_CANCEL, Gtk.ResponseType.REJECT, Gtk.STOCK_OK,
-                Gtk.ResponseType.ACCEPT))
+        creation = True
+        if not board.children:
+            board.children.append("")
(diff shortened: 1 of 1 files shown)


gbtami:
I'v merged this just to see what it's doing, but wouldn't it be a more consistent and quicker solution to add a "Remove comment" menu item to the right click popup menu above "Remove symbols" instead of adding the "Clean" button in to the dialog? This way I don't have to open the dialog then push "Clean" to remove a comment. We already have 3 create menu item and 2 remove. Adding the missing third remove seems reasonable.
Maybe we should change the text "Reset the evaluations" to "Remove evaluations" to be more consistent in wording. What do you think?

ecrucru:
Good evening,

I thought that adding new menu item "Remove comment" would overload the popup menu. I was pretty convinced by the changing caption "Add/Edit comment", so adding a button in the popup looked very simple and unobtrusive imho.

Anyhow, you are right : the item is reasonnable as well as the changed caption to be consistent. Consequently, do you think it is a good idea to group the items as follows ?

- Copy PGN (not totally in the context of the annotation panel but ok)
- Add start comment
- Add comment
- Add move symbol >>
    - ...
- Add evaluation symbol >>
    - ...
- ------------ (separator)
- Remove >>
    - Symbol
    - Comment
    - All the evaluations

In the case you access a comment with a *left* click, keeping the added button makes a little advantage to reduce the clicks if you finally want to delete the comment. That's why I would like to keep that cool button if you agree.

Thanks

gbtami:
Agreed on all points. If you go for it, please remove "Copy PGN" also! It's in the main "Edit" menu and that's more than enough IMO.

PR #1518, opened 2017-12-02
Tags player name and ELO in title

ecrucru:
Hello,

This little PR changes 3 things :
- add ELO in the title, not the tab because the space is limited
- avoid the title "- PyChess" when there is no leading text
- transfer the detected name of the UCI engine to the PGN tags : else the first time you edit the properties of the game, it erases with the name from engineDialog

Regards


gbtami:
Thx!

PR #1525, opened 2017-12-10
Indented export to PGN

ecrucru:
Hello,

This PR adds an optional indentation for the export to PGN. The multi-level analysis is sometimes very complex and once your file is saved, you have everything concatenated and wrapped, so it is not easily readable/changeable in text-mode.

Let's take again a stupid example :

- Before :

```
{Game analysis}
1. e4 e6 2. d4 (2. Nc3 d5 3. exd5 exd5 (3... e5 4. d6 cxd6
(4... Qxd6 {Possible})) 4. Qf3 {Not bad}) 2... d5 3. Nc3 {Several
possibilities} (3. Nd2 c5 4. exd5 $3 Qxd5) (3. a3) 3... h6 *
```

- After :

```
{Game analysis}
1. e4 e6 
2. d4
    (2. Nc3 d5 3. exd5 exd5
        (3... e5 4. d6 cxd6
            (4... Qxd6 {Possible})
        )
    4. Qf3 {Not bad})
2... d5 
3. Nc3 {Several possibilities}
    (3. Nd2 c5 4. exd5 $3 Qxd5)
    (3. a3)
3... h6
*
```

Regards

### DIFF
--- a/glade/PyChess.glade
+++ b/glade/PyChess.glade
@@ -4049,7 +4049,32 @@ It will search positions from http://www.k4it.de/</property>
                         <property name="left_padding">5</property>
                         <child>
                           <object class="GtkCheckButton" id="saveRatingChange">
-                            <property name="label" translatable="yes">Save rating change values</property>
+                            <property name="label" translatable="yes">Save _rating change values</property>
+                            <property name="use_action_appearance">False</property>
+                            <property name="visible">True</property>
+                            <property name="can_focus">True</property>
+                            <property name="receives_default">False</property>
+                            <property name="halign">start</property>
+                            <property name="use_underline">True</property>
+                            <property name="xalign">0</property>
+                            <property name="draw_indicator">True</property>
+                          </object>
+                        </child>
+                      </object>
+                      <packing>
+                        <property name="expand">False</property>
+                        <property name="fill">False</property>
+                        <property name="position">4</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkAlignment" id="alignment29">
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                        <property name="left_padding">5</property>
+                        <child>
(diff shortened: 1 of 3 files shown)


gbtami:
If you want to read .pgn data formatted you have two choice. First, you can use a GUI like PyChess :)
Second, you can create a script based on pychess or python-chess to read->(re)format->save your .pgn

I think this PR is not conform to PGN spec. See http://www.saremba.de/chessgml/standards/pgn/pgn-complete.htm#c3.2

gbtami:
I like to understand the motivation of this PR better. Maybe you just unsatisfied witt the layout of pychess annotation panel. If this is the case, enhancing annotation panel (or movelist panel? like in lichess?) may be the way to go instead of saving .pgn tabulated. Anyway I suggest to open a new issue before you start to implement new things in the future. This way others can comment on your ideas also.

Sorry if I was unfriendly in my prev comment!

ecrucru:
I have not misinterpreted your message. The PR came out of the blue after an unplanned experimental change. It is your right to not accept it.

Another GUI produces an indented PGN (screenshot) but it doesn't support the variants. Its fork has the variants but is not released on Windows. Well, it is not very clear but the basic idea was to add that in pyChess which runs everywhere. For example, if you need to finalize a little change in the file (or undo an incorrect move at sub level for example ;-D), it is easy and very fast to edit an indented PGN file instead of starting a GUI to do that (if possible).

image

You have the choice to opt-in for the indentation in the options. By default, it keeps the existing inline format that you mentioned from the specification. The new option goes a bit against the specification but the resulting file is still loadable without much difficulty in any GUI.

The layout of the annotation panel is excellent and several little issues were corrected recently. There is no concern with it, except the active width of all the widgets (issue 1511).

Regards

gbtami:
Ok I merged it. May I ask you to add a tooltip text to this new config option. Something like "If it's checked it produces formatted (non standard PGN!) .pgn file"?

PR #1528, opened 2017-12-17
Confirm the export to chesspastebin

ecrucru:
Hello,

Its menu item is too close to "Save game as", so it may unexpectly publish data on the Internet without the real consent of the player. The new dialog of confirmation helps to prevent that.

Regards


gbtami:
Thx!

Issue #1519, opened 2017-12-06
NPS display for engine match / Tournament


ghost:
Hello Great team,

is it possible to display the nps for engine and to have a feature to organize tournament between engines ? 
Nonetheless, i really appreciate your software. Very great interface and gameplay.
Thanks
NaltaP

gbtami:
It will be possible if someone implements it :)
Until then I highly suggest to try https://github.com/cutechess/cutechess/releases

PR #1530, opened 2017-12-19
Nodes per second

ecrucru:
Hello,

This PR adds the nps as requested by #1519.
No idea for CECP.

Regards


gbtami:
Seems OK. For missing CECP part see: https://www.gnu.org/software/xboard/engine-intf.html#10
The `anare` regexp of CECP engine at https://github.com/pychess/pychess/blob/master/lib/pychess/Players/CECPEngine.py#L48 needs two additional grouping parentheses around time and nodes parts.

ecrucru:
OK, I understand what you mean. I don't use so much cecp engines, so I would appreciate if you can do that little change in order to validate it more efficiently and quicker. Travis is now OK for the UCI part. Thanks

gbtami:
Ok, no problem. Thx for this PR!

PR #1534, opened 2017-12-28
Redesigned score panel

ecrucru:
Hello,

The current view of the scores in PyChess is vertical, so :
- we have to scroll the view to understand how the score evolves
- we can't take a screenshot of the score
- the informative area between -3/+3 cp is limited in width

This PR suggests a new design :
- the initial black/white style is not modified
- horizontal layout : it will comply with all the existing GUI and websites
- dynamic size of the moves to avoid scrolling : from 1 to 24 px
- linear (-8/+8 cp) and exponential (+max/-min cp) stretching as an option
- horizontal markers for the expectable levels of defeat (±3, ±8 and ±16 cp)

Preview :
image

Thanks for any comment on this.

Regards

### DIFF
--- a/glade/PyChess.glade
+++ b/glade/PyChess.glade
@@ -1643,6 +1643,26 @@
                                       </packing>
                                     </child>
                                     <child>
+                                      <object class="GtkCheckButton" id="scoreLinearScale">
+                                        <property name="label" translatable="yes">Use a linear scale for the score</property>
+                                        <property name="use_action_appearance">False</property>
+                                        <property name="visible">True</property>
+                                        <property name="can_focus">True</property>
+                                        <property name="receives_default">False</property>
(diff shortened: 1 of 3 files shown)


gbtami:
@ecrucru this is awesome! Thx!

Btw. while I tested your PR I noticed a bug (it's in 0.99.1 also, so not related to your work). Jump to the start (Home key) then to the end (End key) of finished game (I used your game from https://github.com/pychess/pychess/issues/1535 ) after some time the light blue highlight bar is jumping to the beginning of the game. Can you confirm this?

ecrucru:
Yes, I have this effect too.

There is another way to see the blue rectangle moving from one position to another. Take a position in a finished game and press the arrow *Right* several times in a row. This will trigger many analyzes but stockfish will not react immediately. So the blue rectangle will change its position (generally backwards compared to the final position) and after some time, it will take the final position. The evaluations are still correct when the effect is active.

gbtami:
OK, filed it as https://github.com/pychess/pychess/issues/1537

PR #1540, opened 2017-12-29
Mate in N

ecrucru:
Hello,

This PR shows a detailed mate in N moves in the bookPanel.

There is no change for CECP but it should be possible to do something similar to UCI.

image

Regards,


gbtami:
Seems good in general. Some notes.
 - flake8 fails with
`lib/pychess/Players/UCIEngine.py:12:1: F401 'pychess.Utils.lutils.ldata.MATE_DEPTH' imported but unused`
 - MATE_DEPTH should be some bigger int value according to http://kirill-kryukov.com/chess/longest-checkmates/longest-checkmates.shtml
 - I suggest to use format_mate as keyword argument with default False
`def prettyPrintScore(s, depth, format_mate=False):`
This seems more readable and needs change in bookPanel.py only.

ecrucru:
I did some adjustments but didn't change the calling parameters for the function as it should not change anymore. Thanks

PR #1542, opened 2017-12-29
Default options in enginesDialog

ecrucru:
Hello,

A button is added below the options to set the default values.
Some engines have a very long list of options, so it may help.

Regards


gbtami:
Thx!

PR #1556, opened 2018-01-09
Should not have double spaces in string

bernardosulzbach:
This is a small visual glitch. I think that my replacement pattern caught everything.

### DIFF
--- a/glade/PyChess.glade
+++ b/glade/PyChess.glade
@@ -1606,7 +1606,7 @@
                                     </child>
                                     <child>
                                       <object class="GtkCheckButton" id="autoPromote">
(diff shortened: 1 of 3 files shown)


gbtami:
@mafagafogigante translators are working on .po files using transifex.com then I pull .po files from transifex.com . They are automatically merged with lang/pychess.pot by transifex when it's changing in pychess git repo. And pychess.pot is generated from .py and .glade files running create_template.po.sh by me. All in all no need to fix them by hand. May I ask you to left them from this PR please?

bernardosulzbach:
I've removed all proposed changes to .pot and .po files.

gbtami:
Cool, thx!

Issue #1529, opened 2017-12-17
Highlight the variation lines


ecrucru:
Hello,

This is a feature request.

The mode human vs human is practical to create studies. For example, in the following position, there are several possible moves :

- `7. Bd2`
- `7. Nc3`
- `7. Kf1`

The request is the ability to show a blue arrow on the board when at least one variation line has been created. It is more user-friendly than checking the annotation panel, and it may serve for quizzes/trainings. Schematically, the output looks like this :

![image](https://user-images.githubusercontent.com/24614488/34079275-0c807b50-e32b-11e7-9cec-4c73c4ac7695.png)

It is similar to #131, but it relies on the variation lines, not on the ouput of the analyzer.

Regards

ecrucru:
Hello,

As a possible alternative, I was thinking about the following design :

![image](https://user-images.githubusercontent.com/24614488/34653187-fa1834e8-f3e8-11e7-8ee6-e998868d198e.png)

- Add moves in little boxes in the header of annotationPanel.
    - If sorted alphabetically, then we can try to guess the best move without assuming that it is always the first one.
    - The layout would be similar to the app «Analyze this» and better that the drop-down list on lichess's studies.

- When the mouse is over the move, an arrow is drawn on the board for a quick read.
    - It will avoid to have N arrows drawn at the same time on the board, as stated in the initial message.

- When you click on the move, you branch to the corresponding main move or sub-move.
    - It will improve the exploration of a multi-level PGN file. That's why the blue boxes of «analyze this» are very practical despite it runs on a mobile phone only and doesn't have sorted moves (guesses are mechanically limited).

- An orange box to go up or to the previous move, not to the main line (aka top level).
    - If you didn't guessed correctly, it will allow to retry.

My naive impression is that driving all the moves from the annotationPanel makes the best sense. Of course, the work _just_ needs to be done and it is not easy.

For educational purposes when reviewing an analyzed PGN file, this feature would be awesome.

Thanks.

gbtami:
Very nice idea! We absolutely need something like this in Learn perspective Lessons panel. Telling the truth the annotation panel code base is rather spaghetti like and needs some refactoring also. First its TextView widget have to be split to header and move text parts because now it scrolls the game header out of visible view if game text is long.

gbtami:
I have started to work on this.

gbtami:
[...]

PR #1557, opened 2018-01-10
Separate header in annotationPanel

ecrucru:
Hello,

A preliminary change as you suggested in #1529. The border/padding may be improved if you have some ideas.

Regards


gbtami:
Thx!

PR #1563, opened 2018-01-15
Fix special chars

ecrucru:
Hello,

The use of `GtkLabel.set_use_markup` implies that the text should be manually escaped.
Else a game `Someone` vs `Me&You` will be shown with readable `<b>` markups in annotationPanel.

Regards

### DIFF
--- a/lib/pychess/perspectives/games/annotationPanel.py
+++ b/lib/pychess/perspectives/games/annotationPanel.py
@@ -1,7 +1,8 @@
-# -*- coding: UTF-8 -*-
+﻿# -*- coding: UTF-8 -*-
 
 import re
 import datetime
+import html
 from math import floor
 
 from gi.repository import Gtk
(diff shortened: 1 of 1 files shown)


gbtami:
Opening a game where players has `WhiteElo` or `BlackElo` from any pgn file I get:
```
  File "/home/tamas/pychess-ecrucru/lib/pychess/perspectives/games/annotationPanel.py", line 1154, in on_players_changed
    self.update()
  File "/home/tamas/pychess-ecrucru/lib/pychess/perspectives/games/annotationPanel.py", line 1122, in update
    self.update_header()
  File "/home/tamas/pychess-ecrucru/lib/pychess/perspectives/games/annotationPanel.py", line 1029, in update_header
    text += " (%s)" % html.escape(elo)
  File "/usr/lib/python3.5/html/__init__.py", line 19, in escape
    s = s.replace("&", "&amp;") # Must be done first!
AttributeError: 'int' object has no attribute 'replace'
```
I agree the need of escaping player names and openings, but I think escaping elo and result tags is unnecessary.

ecrucru:
OK, I removed the other escapes despite the string should be fully escaped to get the correct format. An incorrect status can be injected by pasting a PGN from the menu game, and the ELO can be incorrect from the properties of the game. There is indeed no expectable reason for a player to do this, but this is technically feasible.

The ELO is an `int` or `str` if you load a PGN or start a game from scratch. That's why you found an error. Don't you think that every tag should be stored as a string to avoid this ambiguous effect ?

gbtami:
After taking closer look I'v just realized that your header split commit introduced Labels instead of another TextView, and this causes these formatting issues. I will go back to use TextView instead. Sorry for late response!

PR #1571, opened 2018-01-27
Copy subfen in FilterPanel

ecrucru:
Hello,

The existing button to copy the subfen was hidden in the file `pychess.glade`. It is now implemented as follows :

image

The subfen is above the board to avoid a concentration of buttons in the bottom-right corner of the dialog.

Regards

### DIFF
--- a/glade/PyChess.glade
+++ b/glade/PyChess.glade
@@ -6650,70 +6650,6 @@ If it can use both  you can set here which one you like.</property>
                     <property name="orientation">vertical</property>
                     <property name="spacing">3</property>
                     <child>
-                      <object class="GtkAlignment" id="setup_pattern_dock">
-                        <property name="visible">True</property>
-                        <property name="can_focus">False</property>
-                        <child>
-                          <placeholder/>
-                        </child>
-                      </object>
-                      <packing>
-                        <property name="expand">True</property>
-                        <property name="fill">True</property>
-                        <property name="position">0</property>
-                      </packing>
-                    </child>
-                    <child>
-                      <object class="GtkButtonBox" id="buttonbox1">
-                        <property name="can_focus">False</property>
-                        <property name="layout_style">start</property>
-                        <child>
-                          <object class="GtkButton" id="copy_sub_fen">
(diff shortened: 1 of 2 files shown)


gbtami:
I absolutely forgot about this. I think I used it for testing purpose only when was implementing scoutfish filtering. But you are right, it may be useful for users also. Looking on it now I think using "paste" would be much better naming than "copy" though. What do you think?

ecrucru:
There were 3 buttons : *copy*, *paste* and *clear*.

I principally needed *copy* because the text box is grayed, so it was not possible to extract the subfen made in that window to use it externally.

The button *paste* is removed by lack of reasons to do it. The implementation is a bit longer but possible indeed.

The button *clear* is equivalent to leaving the popup window and retrying. So its removal will save some space.

gbtami:
Ok, I misread your PR code. Sry!

I see the usefulness of adding Paste button here because users may have FEN before they start to search for in a database.
On the other hand we have "Setup position" menu with Copy/Paste/etc. buttons for general use. What is your scenario of using Copy button in this dialog?

ecrucru:
The main interest here is that if you prepare the subfen from the setup dialog, you can't see if it is returning matching results. If the text box was not grayed, no button would be needed to get the subfen.

When I copy the subfen, it is mainly for external purposes, like saving interesting selections, querying scoutfish from another tool, etc... But you are right, the paste button may be needed as well. This is not yet done : I can think about it unless you want do it quicker.

gbtami:
I'm not against (re)enabling sub-FEN Gtk.Entry to be editable and remove all (Cop/Paste/Clear) buttons if you think it will produce more flexible usage of this dialog.

ecrucru:
The purpose is the ability to copy the subfen, that's why the PR initially relied on the reactivation of existing hidden buttons. If activating the edition of the text box may help to copy, paste and clear without buttons, it is also good. No problem.

gbtami:
Merged as is, thx! I will open an issue about adding "Paste". I agree with you that "Clear" has little usefulness here.

PR #1577, opened 2018-02-03
Parse negative increment

ecrucru:
Hello,

While I was adding the estimated duration of a game in the previous PR, I noticed that pyChess offers the possibility to play with a "negative increment" for the local games. That's pretty original and as a logic was implemented to handle the sign of the increment, I assume that it was a deliberate choice.

The problem is that the PGN tag TimeControl is not parseable in that case. With that new PR, "10+-5" will become "10-5" and the UI will be able to parse it and display a nice text in annotationPanel. Indeed, this is not in line with the PGN specification because it requires 2 positives values, but having a negative gain is basically unexpected.

Since the offered gameplay is interesting, I propose to keep this "negative increment" (go against the PGN rules). The Internet part is not changed because the increment is normally positive only.

Regards

### DIFF
--- a/lib/pychess/Savers/pgn.py
+++ b/lib/pychess/Savers/pgn.py
@@ -120,14 +120,14 @@ def parseClockTimeTag(tag):
 def parseTimeControlTag(tag):
     """
     Parses 'TimeControl' PGN header and returns the time and gain the
-    players have on game satrt in seconds
+    players have on game start in seconds
     """
-    match = re.match("(\d+)(?:\+(\d+))?", tag)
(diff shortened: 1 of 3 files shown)


gbtami:
This patch doesn't work for me. Nor parsing "10-5" in pgn.py nor showing it on annotation panel.

ecrucru:
You are right, something went wrong :-((
Here is a new version with a unified regex.
I retested the export to pgn, import from database, and start new game for the 4 cases (300, 300+5, 300-7 and 40/5400).
Thanks

gbtami:
Thx!

PR #1579, opened 2018-02-04
Keep node selected after update

ecrucru:
Hello,

If you apply a NAG symbol on the node which is selected, it will lose its attribute 'selected'.
It is because the node is deleted and recreated. The reapplication of the attribute 'selected' is missing.
The PR fixes that.

Regards


gbtami:
Thx!

PR #1586, opened 2018-02-17
Regular tags

ecrucru:
Hello,

This PR deals with the tags in order that `GameModel.tags` only contains the tags to be exported.

The unconventional tags `Year`, `Month`, `Day`, `Time` and `Result` (see below) are removed, but parsed on the fly to be stored in the database.

All the related hard-coded values/logic are consequently removed as well.

The export of the tag `Result` is now fixed. Because the value `*` was set by default, it was considered as a manual value. Now, you can override the value only if you add it in the tag editor.

When you select the date of the game, the picker is aligned on the mentioned date.

The consequence of all that is that the tag editor is empty when you start a new game, with no specific logic to clean up that view. All the main tags are managed in the dedicated fields on the first tag.

Regards


gbtami:
Great work, thx!
Maybe we can consider to get rid of internal only used Year,Month,Day tags now. They was introduced into game table in sqlite database to simplify processing/presenting incomplete date tags in database perspective. What do you think?

ecrucru:
Sorry, I am not sure to fully understand your question.

In the current idea, `GameModel` would only have the tag `Date` that you can split with `getGameDate` : it supports `2000.??.01` for example in respect of the specification. In the unchanged PgnImport, it splits it into nullable integer fields (little change to come with the next paragraph). That logic may continue imho and we may foresee something new to filter by month as well.

The PGN saver needs other changes : this is part of the remaining tasks of the already opened ticket. This PR gave me some ideas to continue on it.

I have not validated the tags in regards of ICS games. They may still produce the extra tags removed here.

For the usage we have of the tag `Result`, it is a bit redundant with `GameModel.status`, so I see no reason to keep the initial value `*` which causes a problem in the recent dynamic export to PGN. I have not observed any particular issue yet, but I stay careful if you notice something.

gbtami:
I see very little value filtering .pgn by Month or by Day. I like to enhance Date filtering by prefix search instead similar to names. For example: 2018 or 2018.02 or 2018.02.19

Issue #1583, opened 2018-02-07
Can't use Learn without Stockfish


ecrucru:
Hello,

When Stockfish is not installed, it is not possible to start a puzzle from the home screen. Could PyChess use the default analyzer set in the preferences if it supports the standard chess ?

In my example, I replaced the default package Stockfish by ddugovic/stockfish to cover most of the variants. Other players may prefer a commercial engine. Etc...

Thanks

gbtami:
PyChess Linux packages (.deb .rpm) depends on stockfish from 0.99.1 so if you install pychess it will install stockfish also. (On Windows .msi is boundled with stockfish.)

Using the default analyzer set in preferences seems a bad idea because it even can be pychess own engine which is so weak and slow it may not find mate in x some(?) cases.

gbtami:
I myself use ddugovic/stockfish also. Maybe we should add ddugovic/stockfish to externals download dialogue until it's not in big Linux distributions.

ecrucru:
I agree with you concerning the capabilities of pychess engine. The concern is more about finding a suitable engine dynamically for the puzzles to not break Learn as above. There is no visible error so it may be weird for chess players. The logical procedure could be :
- preference if it is `*stockfish*` (in case of multiple instances of stockfish)
- find a standard `*stockfish*` in the list of engines (it matches against `stockfish8.exe` and `stockfish_8_xBIT.exe`)
- pick the prefered engine
- find the first classical engine in the list (if prefered one is a pure variant engine)
- nothing found

By the way, Stockfish 9 has been released this week.

Yes, adding ddugovic/stockfish to the external downloads is a nice idea, provided that suitable binary files are available. It was not the case until recently.

gbtami:
I agree that finding any version of stockfish should be enhanced. On the other hand PyChess packages (.deb .rpm .msi) will always install stockfish so the concern is valid only in source install cases. Implementing a pop up dialogue to inform the users when no stockfish version is installed when they wants to start a puzzle seems enough for me.

PR #1597, opened 2018-02-20
Learn without Stockfish

ecrucru:
Hello,

For the view Learn, this PR finds another engine when Stockfish is not available for the reasons mentioned in #1583. It stays the primary engine to be found.

This change will be more resilient for PyChess on Windows than the current logic.

Regards


gbtami:
Thx!

Issue #1610, opened 2018-03-01
Deeper annotationPanel


ecrucru:
Example to illustrate the idea :
![image](https://user-images.githubusercontent.com/24614488/36863640-939bc0d8-1d8a-11e8-98b6-f446461a06bb.png)

gbtami:
Can you explain this a bit more detailed please?

ecrucru:
The levels are not well indented because of a limit given by the red style.

I have some PGN studies with a lot of levels and I just realized that a deeper level may be less indented than the parent level. With wrapped text when the subline is long and/or commented, it becomes very complicated to handle.

The idea is to see if adding new styles may help to limit that effect. It will not change anything until 3 sublevels.

PR #1612, opened 2018-03-02
Deeper annotationPanel #1610

ecrucru:
Hello,

Here is the new suggested design for annotationPanel with enhanced alignments and colors in the limit of 64 levels :

image

It is more good looking than :

image

The PGN to reproduce the case is :

```
1. e4 e5 
2. d4
    (2. Nf3 Nf6
        (2... d6 3. Nc3 Nf6
            (3... d5 4. Nxd5 c6 5. Nb4
                (5. Nc3 Qd4 6. Nxd4 exd4 7. Nd5
                    (7. Ne2 d3 8. Nd4
                        (8. f3 dxe2 9. Qxe2
                            (9. Kxe2 Be6 10. c4
                                (10. b3 Nd7 11. c4
                                    (11. d4 c5 12. dxc5
                                        (12. d5 Bxd5 13. Qxd5)
                                    12... Nxc5)
                                11... Ngf6)
                            10... Bxc4+)
                        )
                    )
                )
            5... c5)
        4. Nd5)
    3. Nc3 d6)
2... f6 *
```

Regards

### DIFF
--- a/lib/pychess/perspectives/games/annotationPanel.py
+++ b/lib/pychess/perspectives/games/annotationPanel.py
@@ -43,8 +43,7 @@ def add_provider(widget):
     style = widget.get_style_context()
     provider = Gtk.CssProvider()
     provider.load_from_data(css.encode('utf-8'))
-    style.add_provider_for_screen(screen, provider,
-                                  Gtk.STYLE_PROVIDER_PRIORITY_USER)
+    style.add_provider_for_screen(screen, provider, Gtk.STYLE_PROVIDER_PRIORITY_USER)
 
 
 # -- Documentation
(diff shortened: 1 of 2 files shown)


gbtami:
Great patch! I have two minor notes.
1. Can you get rid of square brackets and use parenthesis in every level of variants please? It was my idea to use the former in main variations and the latter for all other ones, but I think this has no meaning after fixing this indentation issue.
2. Some of the users of PyChess uses dark desktop themes and some of your predefined colour lines are rather hard to see. I used to test this with Arc-Dark Theme on Ubuntu 16.04. See http://www.noobslab.com/2017/01/arc-theme-light-dark-versions-and-arc.html

ecrucru:
OK, the brackets are removed.

For the background, I have no installed dark theme. Could you confirm that the previous green and red styles were fitting both with clear and dark themes ?

There is no existing logic to handle a dark theme. Should we calculate the lightness (HSL) of the default bg_active to know automatically that it is dark or light ? Do you know a magic function ? I can already check for a new palette of colors for a dark theme.

gbtami:
Unfortunately I have no answer to your questions. All I can suggest is to install Arc theme from https://github.com/horst3180/Arc-theme and test PyChess using it.

gbtami:
Old-new creenshots using Arc-dark theme:
new
old

ecrucru:
The theme finally worked on the desktop but not with PyChess :-(
I tested with some workarounds, the new commit should be able to detect the black theme.
May I ask you to test please ?

gbtami:
Seems better now. Thx!

PR #1617, opened 2018-03-10
Refresh annotationPanel

ecrucru:
Hello,

When you prepare a study in pychess, you can move the pieces at various levels and order. The problem after some time is that it misaligns the order, the style and the hierarchy of the elements of the textarea in a way that only a full refresh can restore the view correctly.

The ability to refresh the view by playing around with the comments is no more possible since 626d0feb3da38b8129aee359e405f462888c777d. The idea is to add a popup item "Refresh" in annotationPanel, else the study cannot be completed in correct conditions.

Regards

### DIFF
--- a/lib/pychess/perspectives/games/annotationPanel.py
+++ b/lib/pychess/perspectives/games/annotationPanel.py
@@ -344,6 +344,10 @@ class Sidepanel:
                         position = index
                         break
 
+                menuitem = Gtk.MenuItem(_("Refresh"))
+                menuitem.connect('activate', self.menu_refresh)
+                self.menu.append(menuitem)
+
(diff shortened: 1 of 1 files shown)


gbtami:
"misaligns the order, the style and the hierarchy of the elements of the textarea"
If this is a bug in pychess adding a refresh button doesn't feels like a correct solution for me. May I suggest to create an issue with detailed description of problem instead, please?

ecrucru:
Yes, I agree with you. As the issue is not easily reproducible, it is a workaround that existed indirectly and that can be removed in the future. The issue occurs regularly with 0.99.1 and the full refresh is essential.

gbtami:
I'm still thinking that fixing the original bugs should be better, but having this workaround is better than not having it :)

PR #1619, opened 2018-03-12
Translatable header TC

ecrucru:
Hello,

`TC` means nothing in non-English language.
This PR allows to have a translation for this column.

Regards


gbtami:
Thx!

PR #1626, opened 2018-03-26
Signed mate in

ecrucru:
Hello,

The idea is to show if the displayed mate in is advantageous for white or black.
There is a color in bookPanel but not in annotationPanel.
Sometimes, the side of the mate in may change in 1 move.
This PR fixes that : `Mate #-2` or `Mate #3`

Regards


gbtami:
Thx!

PR #1628, opened 2018-03-31
Corrected link of Maurizio Monge

mliebelt:
The old link did not reach the home page, after finding the correct one, it should be added in artists as well.


gbtami:
Thx!

Issue #1629, opened 2018-03-31
Enhanced load of the local engines


ecrucru:
Hello,

The concept is to get more information about the engines based on matching names, not based on hard-coded names.

The work in progress includes :
- a reworked list of engines (to be extended later)
- insentiveness to exact file name, including :
    - software version
    - operating system
    - bitness
    - instruction set
- a new field ELO

Regards

gbtami:
Can you explain a bit more what is the goal of this issue please?
Or we can discuss it on http://webchat.freenode.net/?channels=pychess&uio=d4 if you have time.

ecrucru:
It is just minor things to investigate to get a proposed flag and ELO estimate against an engine whatever the exact name of the executable file. In the same way, it should fix little things I saw like hard-coded values, clean up or duplicate engines.

It should help me also to qualify upcoming harder issues for 6 engines of various strength and quality (Andscacs, Gull, µMax, Phalanx, Vanilla and AnMon).

Issue #1574, opened 2018-01-30
Highlight the modified options of the engines


ecrucru:
Hello,

This is a feature request.

Taking in example Firefox, a modified option is put in bold text. Can we do something similar for the options of the engines ?

Komodo, Rybka or Toga2 (for example) have many options and it is not always easy to see which ones were modified manually.

Regards

PR #1632, opened 2018-04-03
Enhanced management of the local engines #1629

ecrucru:
Hello,

The changes are wider than initially expected :

- new field "Comment" for the engines to store a free text
- new field "Preferred level"
	- adjustment of the slider when the engine is selected for a new local game
- detection of the protocol, flag and ELO based on a to-be-enhanced list of engines
	- ordered lookup XB/UCI when the engine is added
	- `stockfish_9.0_32bits.exe` and `stockfish9_x64_popcnt` will return the same flag and ELO
- engines saved when you close the dialog
	- ability to undo the current changes
	- popup to confirm in case you leave without saving
- automatic cleanup of the expired engines with adjustment of the favorite analyzers
- forbidden duplicate engines which names should be unique
	- check added when you add it : no more use of "name(1)" (note: it was not cascaded to "name(n)")
	- removed dead code linked to the ability to rename
	- removed hidden button "copy engine"
- added mark for the modified options (#1574)
- few internal cleanup and removal of hard-coded values

Regards


gbtami:
1. Unfortunately I can't test this PR because I get:
```
tamas@tami:~/pychess-ecrucru$ ./pychess
Warning: eco.db not found, please run pgn2ecodb.sh
  File "/home/tamas/pychess-ecrucru/lib/pychess/Main.py", line 485, in do_activate
    dd = DiscovererDialog(discoverer)
  File "/home/tamas/pychess-ecrucru/lib/pychess/widgets/discovererDialog.py", line 24, in __init__
    discoverer.pre_discover()
  File "/home/tamas/pychess-ecrucru/lib/pychess/Players/engineNest.py", line 277, in pre_discover
    self.engines = [engine for engine in self.engines if os.path.isfile(engine.get('command'))]
  File "/home/tamas/pychess-ecrucru/lib/pychess/Players/engineNest.py", line 277, in <listcomp>
    self.engines = [engine for engine in self.engines if os.path.isfile(engine.get('command'))]
  File "/usr/lib/python3.5/genericpath.py", line 30, in isfile
    st = os.stat(path)
TypeError: stat: can't specify None for path argument
```

2. Travis shows some failed tests also.

3. I'm not convinced about adding hard coded Elo to engines is a good idea. On the one part Elo is not an absolute value, it depends on several things (hardware, time control, pool of games played, etc.-etc.) and continuously changes on the other part if we pick one list (for exaple CCRL 40/40) there never will be any guarantees that a particular engine of user has installed is the same version listed on the list we used. Some Linux distributions can have very ancient versions of different engines.
Maybe grouping engines to amateur,master,gm or something like this could be a better idea.

4. Seems this PR changes lot of things at once. May i ask you to split it to smaller independent PRs please?

ecrucru:
Well, the build is now completed : you and Travis had a different initial situation causing some interesting unhandled troubles. It is now corrected as required.

For the ELO, like any rating (even in sport), it is necessarily a function of many parameters. What matters here is the scale in order to prepare relevant AI-AI games to not exceed a difference of 400 ELO. There is also another interest mentioned in the updated first message.

Your idea to qualify the value into a readable group is nice. Should you add an issue for that with all the details you have in mind ?

Do you know why PyChess is not listed in the CCRL/404 ? That could be an interesting information.

I would not recommended to split the PR. Your general feedback is rather appreciated to make any further correction if needed. I have no bug in mind about this PR.

Thanks

gbtami:
First of all thx for the correction! I can test this PR now.

1) My arguments against adding hard coded engine Elo values to pychess still stands. I don't want to maintain and ever growing and changing list of engines and their randomly chosen Elo which doesn't match the engines version installed by the users. The initial goal of adding hard coded engines listed in pychess was only to help auto recognise installed (from distribution packages!) engines for Linux users. No more no less. For all other engines we provide the "Add" on engines dialog.
(You can say that this list today contains some windows .exe engines etc-etc, but these are exceptions added by me to easy testing wine working and similar.)

The "use of the declared ELO attribute to search for a learn engine having at least 2500" use case is invalid here, because stockfish is a hard dependency for pychess because it's 1.one of the strongest 2.open source 3. available in all big/popular Linux distribution 4. it has multi PV UCI option!

This latest is important and causes me to NOT able to accept "consideration of the the XB attribute analyze for the pick of the learn engine (note: you could only use UCI engines before)" in this PR also.
XB engines has no support for multi PV analysis mode which is essential in learning module because pychess uses it to provide 2 best moves into hints needed to accept both Qxe8 and Rxe8 in https://github.com/pychess/pychess/blob/master/learn/puzzles/lichess_study_lichess-practice-checkmate-patterns-i_by_arex_2017.01.22.pgn#L18

2) Regarding new field "Preferred level"  I don't understand why it is better to use 10 as very first time default level instead of 20. I feel this only makes the GUI cluttered without any gain.

3) "consideration of the the XB attribute analyze for the pick of the learn engine (note: you could only use UCI engines before)"
As I mentioned above picking of learn engine should search only for any installed version of stockfish.

4) "forbidden duplicate engines which names should be unique "
I better like to see a patch that enables adding same engine with different names. This can be useful to support Rodent III personalities (http://talkchess.com/forum/viewtopic.php?t=61837) or just adding the same engine with different default settings (like ponder on/off and similar).

ecrucru:
1. Ok, I will replace it with a free comment. However I would be interested to know why in the code "you" use is_analyzer with a specific rule for XB ? Initially, the exclusion of XB was to exclude pychess as you suggested it was not strong enough.

2. The default level is changeable easily. Ok for 20.

3. see 1

4. This is not part of the planned changes but it is a nice thing to have indeed. We had previously an issue causing the engines not to be renamed, so they can't be duplicate for now.

Thanks

gbtami:
"why in the code "you" use is_analyzer with a specific rule for XB "
This is because "analyze" feature is optional for CECP protocol engines while "go" is required by UCI.

ecrucru:
Here is the update and the first message is updated to consider that discussion.

gbtami:
Thx for the update!

Issue #1642, opened 2018-04-21
Add the engines in mass


ecrucru:
Hello,

In the current state, the engines are found automatically by looping on a preconfigured set of engines, then the run data are found in the path. An engine can be later added manually from the dialog window.

The feature request is about adding the engines by selecting a folder. Indeed, they are generally stored in the same parent folder.

So the idea is :

- Add a function(path, depth)
	- If path is none, consider the envpath
	- Scans the folders to match the depth
	- Return a list of folders
- At startup, we call the function for the envpath and depth 0
- In engineDialog, add a button
	- The button selects a folder from the dialog
	- A list a subfolders is returned : it should not be part of envpath
- For both, find an executable file in each folder : `*.exe` or `+x`
- If the file matches against a pattern like `uninstall|reset|remove|delete|purge`, skip
- If the file is already loaded (same name), skip to not add it twice
	- If you add again the same folder, it will do a delta load
- If the file is not in the list of preconf engine list, skip (else we could trigger an executable file not related to chess)
	- It is still possible to add it manually
- If the file name contains `32`, `64`, etc... we verify that the bitness matches with the operating system
	- The engines are often shipped in multiple copies
- Add the engine in a to-be-checked state (like today)
- Once the loop is finished, PyChess will use the existing feature to load the options
- We need then to clean the engines for which nothing was read

Regards

gbtami:
Good idea! I always hate to re add my test engines after deleting engines.json for any reason.

PR #1643, opened 2018-04-21
Extended list of engines

ecrucru:
Hello,

Here is the extended list of reliable engines that can be detected.
It provides accurate default values, and it may serve for the enhancement #1642.

Regards


gbtami:
As I said before I don't want to maintain an ever growing and changing list of engines in pychess.

gbtami:
@ecrucru First of all, sorry if I seemed too negative. As I think today my only concern regarding this PR is slow startup time of PyChess. But if this PR is just to help implement #1642 and doesn't affect auto querying distribution installed engines on pychess startup, I'm fine with it.
Maybe after stockfish being official dependency of pychess, we should drop checking distribution installed engines (at least the part what is based on our hardwired list, but preserve HGM method!) on pychess startup. What do you think?

ecrucru:
Hello, I am happy that you finally realized the indirect interest of that PR. I still use it personally from the time it was posted and I don't see any issue with it. It might need some optimization indeed.

To handle the adding in mass of the engines, the checklist could be :
- the default auto-detect with full list is done once at start-up by config
- the default auto-detect is trigerable from engineDialog : it will scan the environment PATH, plus some hard paths like `/usr/bin/games`
- add in mass with full list from new "Add folder" : it will not scan PATH but 2 or 3 sub-levels of the selected folder
- the find of shutil should be replaced by something supporting wildcards, bitness, file attribute, pattern of exclusion for file names, cpuid, paths out of/in complement of the environment PATH... That's hard, please refer to the other issue
- adapt the unit tests so that travis detects the file "stockfish_9_x32.exe" by just giving "stockfish"
- check if the HGM method is still applicable

Theoretically, the list works both for Windows and Linux if we don't specify something like "stockfish_9_x32.exe".

gbtami:
If you still interested on #1642 may I ask you to re send this PR, please?

PR #1660, opened 2018-08-04
Fix value of K in ELO estimation

ecrucru:
Hello

A game cannot be considered as blitz if the time control is unknown. This patch will principally provide a more accurate estimation of the elo variation for grand masters (K=10 if >=2400). To be checked in the properties of the game.

Regards


gbtami:
Thx!

PR #1678, opened 2018-09-22
Extended list of engines (update)

ecrucru:
Update for #1643


gbtami:
Thx!
I removed some engines to let unit test pass on travis.

PR #1682, opened 2018-09-28
Updated tips of the day

ecrucru:
Hello,

Following the todo "Games", this is an update for the tips of the day.

Before, we showed the messages randomly if the popup had to be shown. Now, we have a new list managed by category, shuffled in only 1 possible way at runtime so that the messages cannot be displayed several times in a row, even after restart. It should be user-friendly.

HaGN


gbtami:
Cool, thx!

PR #1696, opened 2018-10-18
Configurable auto-detection of the engines

ecrucru:
Hello,

The engines which name conflicted with a Linux package are reactivated but excluded from the automatic detection. The interest is keeping the default information of the engine for a manual setup.

The engines below 2700 are not detected automatically. The packaged engines are generally very strong. The detection based on the list would be more promising with the logic described in #1642.

The automatic detection is done at every startup. Should we only do it once, but keep the update at every startup ? Can the automatic detection at startup be an option in the preferences, or a button in engineDialog ? The installed engines don't change every day, so the idea is to reduce the frequency of that detection.

Thanks for your feedback.


gbtami:
The original idea behind auto detection comes from Linux use case. When a user installs/removes a chess engine package of hes distribution we want it to be appear/disappear automatically in PyChess.
Linux distributions has only couple of chess engines packaged (and seem to add no new ones last years), so originally our list was rather short and fixed length.
I like to keep this strategy and use this new extended list only for #1642 to prevent growing pychess startup time more (it's already rather slow on Windows).

PR #1711, opened 2018-12-01
Support for UCI_AnalyseMode

ecrucru:
Hello,

This PR adds the support for that UCI option, available in SF10 for example.

Have a nice weekend.


gbtami:
Thx!

PR #1721, opened 2018-12-15
Find engines from path

ecrucru:
Hello,

For the enhancement #1642, this PR adds the function that will scan the folders recursively to find a referenced engine.

The function is alone yet. The idea later is to add a button in enginesDialog to request a root folder. This function is then called. The returned list of files is shown in a popup. The player ticks the relevant engines. The selected one are loaded if they don't exist in `engines.json` yet.

That's how Arena works and I don't see how we can do it differently. We need the verification of the player so that we don't execute invalid executable files which may generate errors into PyChess.

Regards

### DIFF
--- a/lib/pychess/Players/engineList.py
+++ b/lib/pychess/Players/engineList.py
@@ -8,19 +8,13 @@ from collections import namedtuple
 AUTO_DETECT = True
 NO_AUTO_DETECT = False
 
-PYTHONBIN = sys.executable.split("/")[-1]
-
+# CPUID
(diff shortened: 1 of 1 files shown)


gbtami:
Wouldn't it be more elegant using os.walk() instead?

ecrucru:
Hello gbtami. I initially wanted to limit the depth of the exploration in case one use a high-level folder. By simplification, I have not written it yet and `os.walk()` will probably not help much for that purpose. Probably there is a function for that (I cannot confirm yet).

gbtami:
OK, np.

PR #1725, opened 2018-12-21
Redirect files in drag&drop

ecrucru:
Hello,

With that PR, if you drag and drop a file SCOUT/BIN/SQLITE, it will load the PGN file instead.

HaGN


gbtami:
Thx!

PR #1735, opened 2019-01-16
Limited depth and symbolic links

ecrucru:
Hello,

For the mass detection of the engines, this PR adds the control of the depth and the symbolic links.

If you select a big folder, it will be faster and safer.

HaGN


gbtami:
Thx!

PR #1741, opened 2019-02-24
Use Scoutfish in direct database query

ecrucru:
Hello,

The following button relies on ChessDB to check if the database contains a valid position. But when the database is very big, the generation of the BIN file may be impossible. So the button becomes useless and tells that no position can be found.
image

This is partly incorrect because you can filter manually by Scoutfish that handles lighter files. But it requires many clicks for the player. Technically speaking, the approach is not strictly the same, yes.

Adding Scoutfish to the query is the topic of that PR. I appreciate your comments if you see further possible improvements.

Thanks


gbtami:
I think this is a good enhancement, thx!

Issue #1690, opened 2018-10-08
Open URL


ecrucru:
Hello,

The feature would be the ability to open a URL into PyChess to get one or several games that can be analyzed and commented locally without doing all the steps (download, select file, load to database, clean .bin/.scout/.sqlite at the end, etc...). If there is only 1 game, then it is loaded directly without database as if we clicked on "_Enter game notation_". Sharing links are common by mail.

- [x] lichess.org/{ID} --> lichess.org/game/export/{ID}?literate=1
- [x] lichess.org/study/{ID} --> lichess.org/study/{ID}.pgn --> database
- [x] chessgames.com/perl/chessgame?gid={ID} --> chessgames.com/pgn/anyfilename.pgn?gid={ID}
- [x] chess.com/live/game/{ID} --> parse
- [x] theweekinchess.com/twic --> already exists
- [x] ficsgames.org/cgi-bin/show.cgi?ID={ID} --> ficsgames.org/cgi-bin/show.cgi?ID={ID};action=save
- [x] chess24.com/en/game/{ID} --> chess24.com/en/download-game/{ID} <-- gate
- [x] chesstempo.com/gamedb/game/{ID} --> chesstempo.com/requests/download_game_pgn.php?gameids={ID}
- [x] chesspastebin.com/* --> parse encoded HTML link
- [x] chess.org/play/{ID} --> parse encoded HTML link
- [x] 365chess.com/view_game.php?g={ID} --> 365chess.com/download.php?guid={ID} <-- gate
- [x] europe-echecs.com/* --> parse links europe-echecs.com/*.pgn --> database
- [x] link of mime `application/x-chess-pgn` --> load
- [ ] link of mime `application/zip` --> unzip --> database
- [x] generic parse

The link can be opened by dragging and dropping a .url file onto PyChess.

That's all for the recent ideas.

Thanks

PR #1748, opened 2019-04-12
Open URL

ecrucru:
Hello,

This PR solves #1690. By giving the URL, you can load a finished remote game into PyChess as a new game.

Can you help to regenerate the POT file after ?

Thanks.

### DIFF
--- a/glade/PyChess.glade
+++ b/glade/PyChess.glade
@@ -160,6 +160,11 @@
     <property name="can_focus">False</property>
     <property name="icon_name">x-office-calendar</property>
   </object>
+  <object class="GtkImage" id="image27">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="stock">gtk-jump-to</property>
+  </object>
   <object class="GtkImage" id="image3">
     <property name="visible">True</property>
     <property name="can_focus">False</property>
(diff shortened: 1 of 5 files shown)


gbtami:
Doesn't work for me with https://lichess.org/study/hr4H7sOB.pgn

gbtami:
For me unit test produces 1 error.

ChessTempo.com -- Download link
- Target link : http://chesstempo.com
- Expecting data : False
- Fetched data : True
F
======================================================================
FAIL: test (__main__.RemoteGameTestCase)
----------------------------------------------------------------------
  File "remotegame.py", line 24, in test
    self.assertEqual(ok, expected)
AssertionError: True != False

----------------------------------------------------------------------
Ran 1 test in 2.023s

FAILED (failures=1)

gbtami:
All test data (all get_test_links()) belongs to unittest. Can you move them there, please?

ecrucru:
Hello,

Lichess studies were not implemented yet.

For ChessTempo, a precondition failed out of the class due to a wrong piece of code. The test case is adjusted accordingly.

For `unittest`, the URL to fetch depend on the classes activated in `chess_providers`. So it doesn't make sense to me to put the test URL out of their classes. Do you see another solution ?

HaGN

gbtami:
You can make individual test cases for different chess_providers something like movegen tests here https://github.com/pychess/pychess/blob/master/testing/movegen.py#L151

If you like extreme data driven approach you can go for something like this https://github.com/mcostalba/scoutfish/blob/master/src/test.py

While the latter seems more elegant, the former is much simpler and fits better to Zen of Python: Explicit is better then implicit :)

gbtami:
Cool, thx a lot!

PR #1754, opened 2019-04-25
Fix chess.org under Linux

ecrucru:
Still checking things...

### DIFF
--- a/lib/pychess/Savers/remotegame.py
+++ b/lib/pychess/Savers/remotegame.py
@@ -137,6 +137,7 @@ class InternetGameInterface:
         loop = asyncio.new_event_loop()
         asyncio.set_event_loop(loop)
         result = loop.run_until_complete(coro)
+        loop.close()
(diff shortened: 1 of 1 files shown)


gbtami:
Thx!

ecrucru:
Hello gbtami,

I again not totally finished the correction, but ok, it should solve the issue in the background for now. Travis' runs picked the 2 links that don't produce any data, so it proves nothing yet. I retested in a VM, so Travis is expected to be happy, not me.

The problem of integration for ChessOrg is that PyChess GUI is running its own event loop. If you open the link from the GUI, it will still not work under Linux. Python says that the event loop is already running, so I cannot run anything on top/place of it from the method `async_from_sync`.

I run out of ideas... and there is no problem under Windows. That's why I have not seen that earlier.

Thanks for any tip regarding asyncio and event loops.

gbtami:
You can try with synchronous https://github.com/websocket-client/websocket-client

PR #1755, opened 2019-04-28
Drag and drop .URL files

ecrucru:
Hello,

With this PR, you can drag and drop the files having the extension `.url`.

Regards

### DIFF
--- a/lib/pychess/Main.py
+++ b/lib/pychess/Main.py
@@ -9,7 +9,6 @@ import platform
 import sys
 import subprocess
 from urllib.request import url2pathname, pathname2url
-from io import StringIO
 
 from gi.repository import Gdk
 from gi.repository import Gio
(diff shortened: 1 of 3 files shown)


gbtami:
Where you get these .URL files from? I just like to try this PR with something :)

ecrucru:
Just drag and drop a favorite link from your webbrowser onto your desktop, and you get a file .URL supported by all the known browsers. It is practical format to store a link that you can open in your browser with a double-click.

It has the following basic content :

```
[InternetShortcut]
URL=https://lichess.org/training/84969
```

gbtami:
Thx. I'm not using any icons on my Linux desktop from many years, so I'v completely forgot about them :) After enabling this again, seems they are .desktop files on Linux, containing something like this:
```
[Desktop Entry]
Version=1.0
Type=Link
Name=https://lichess.org/training/84969
Comment=
Icon=user-bookmarks
URL=https://lichess.org/training/84969
```
May I ask you to add support for these also, please?

ecrucru:
Indeed, here it is :)

I also refactored a bit the handling of the error popup because you can drag and drop several files at once.

gbtami:
Thx!

PR #1763, opened 2019-07-18
Collection of changes

ecrucru:
Hello,

Here is a collection of new little changes :

- The option `UCI_LimitStrength` will not be used in analysis mode. The project Stockfish explains in the PR 2225 how its new option will work soon. So if you reduce the strength for your games against AI, it should not be applied to the analyzer.

- The option `UCI_Elo` will be bounded between the defined min/max when a new game is started. To date, the value is 150 times the strength. However, it produces unsupported values. In the PR mentioned above, Stockfish will theoretically not run under 1350. It will convert the ELO to a fractional Skill Level, and the formula would not accept 750 ELO for example. The formula in PyChess is corrected as follows : level 1 = min UCI_Elo, level 18 = max UCI_Elo proportionally to 18/20. If the min/max are not detected, it will use a default range of 1000-2800, so 100 ELO per strength step. An old free version of Rybka has the same kind of limitation, and based on PyChess log, the UCI value looks more consistent now.

- Various typos.

- The exceptions in remotegame are more explicit by changing their representation.

- Broadcasts from Lichess can be opened as studies. The unit test is updated accordingly.

- Unknown ELO won't be displayed in the main application title. Indeed, if you load a PGN having `[WhiteElo "?"]`, you had before a useless `(?)` in the title.

Thanks for your comments

### DIFF
--- a/lib/pychess/Players/UCIEngine.py
+++ b/lib/pychess/Players/UCIEngine.py
@@ -96,6 +96,8 @@ class UCIEngine(ProtocolEngine):
         if analyze_mode:
             if self.hasOption("Ponder"):
                 self.setOption('Ponder', False)
+            if self.hasOption("UCI_LimitStrength"):
+                self.setOption('UCI_LimitStrength', False)
         if self.hasOption("UCI_AnalyseMode"):
             self.setOption("UCI_AnalyseMode", analyze_mode)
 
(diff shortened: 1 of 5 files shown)


gbtami:
Great, thx a lot!

PR #1765, opened 2019-07-21
Drop chess website

ecrucru:
Hello,

One of the latest build failed after the upgrade of this site.
The live games are now encoded like the daily games.
This PR is a drop of support to keep pychess buildable.

Regards


gbtami:
Thx!

PR #1766, opened 2019-08-01
Improve grammar on externals dialog

bboutkov:
Hi. Please find in this PR a small grammar improvement to the externals dialog. 

I hope that I found all the files that need to be updated. Please let me know if this is OK or if anything else needs to be updated!


bboutkov:
I'm not exactly sure how the coverage could change with this commit as it was basically a `find . -type f -exec sed ....` commit, so I expected the coverage to remain static. Any hints are appreciated if this needs to be updated further!

gbtami:
Coverage is not that important :)
But the PR is unnecessary excessive. Changing in .py only is enough because *.po and.pot files are auto generated from source files.

gbtami:
Thx!

PR #1770, opened 2019-08-24
Collection of changes

ecrucru:
Hello,

This PR contains few little changes :
- the detection of popcnt used in `pgn` (Linux only) is now usable for the mass detection of the engines
- technical code improvements in `remotegame`
- open the remote practices of lichess
- when you opened a pgn file, you had plenty of console errors from the analyzer due to conversions with `int`
- the unknown ELO should not be shown in `annotationPanel` (that's similar to what was done for the main title)

Happy week-end

### DIFF
--- a/lib/pychess/Players/engineList.py
+++ b/lib/pychess/Players/engineList.py
@@ -1,7 +1,7 @@
 import os
-import platform
 import sys
 from collections import namedtuple
+from pychess.System.cpu import get_cpu
 
 
 # Constants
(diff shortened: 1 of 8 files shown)


gbtami:
Just notice that fishnet uses detect_cpu_capabilities() to decide which stockfish executable should be download for users. Maybe we should borrow that.
https://github.com/niklasf/fishnet/blob/master/fishnet.py#L961

ecrucru:
There are several techniques for the detection and most of them work on Linux only.
On Windows, there is [assembly code in Python](https://gist.github.com/niklasf/a33a7405f88fbc7a27f0bd6f33a73678) : a matter of taste...
The existing logic is fine and very simple. No reengineering was made on this topic.
Thanks for the merge :)

gbtami:
Thx for your work! :)

Issue #1411, opened 2017-03-18
Predefined positions


ecrucru:
Hello,

This is a feature request.

What about adding a button in the board editor to set up the classical position ? Another one to clear the board ? Possibly amother one to set up Chess960 ?

For now, we have to start a new game before going to the board editor for the first time, then we get the default position.

Thanks

gbtami:
One remaining enhancement is to add button to set up Chess960 start positions.

gbtami:
I think adding another button to set up Chess960 is just clutters the GUI without real need.

ecrucru:
On this topic, let's generate the 960 positions in `eco.db`, then the position can be called by name through the existing button.

PR #1774, opened 2019-09-03
Enhanced opening book

ecrucru:
Hello,

This PR contains many enhancements for the opening book :

- new opening book made with >780k sanitized games from TWIC, having more than 20 moves, players with >1800 ELO, from 2013 to 2018, restricted to 26 plies max, each position being played at least 10 times
- new English ECO file from nikslaf/eco (book coverage of 99.5%)
- ECO names translated into French
- show the opening in English when the translation doesn't exist
- new db structure for `eco.db` with all the unique positions of the openings
- in bookpanel, if a move has no name, the shortest and deeper name is proposed
	- the option can be turned off in the preferences
	- it is disabled by default
- new feature : start from an ECO code or any matching description (like `pianissimo`) by clicking on "Initial position" (familiar with #1411)
- fix the determination of the openings when undo is called :
	- play `e4`, undo : the opening is now reset
	- play `e4 e6 d4 d5 Nc3 (Paulsen) Nc6 exd5 (Svenonius)`, undo : the opening is redetermined at ply N-2 Paulsen, so Svenonius at ply N doesn't remain because the ply N-1 has no name
- when a game is loaded, the opening is not redetermined if the game has the tags `ECO` or `Opening` defined already

Thanks for your review.


gbtami:
Seems unit tests fails.

ecrucru:
Well, [this line](https://github.com/pychess/pychess/blob/master/testing/dialogs.py#L93) is now calling a modal dialog to get a user input. The unit tests are out of the pychess library. So do you know if there is a good manner to detect that we are running a unit test ?

gbtami:
Thx for fixing it!
I'm sorry for not responding before.

gbtami:
Hmm. After merge I get this error when I start a new game against Stockfish
```
  File "/home/tamas/pychess/lib/pychess/perspectives/games/annotationPanel.py", line 1272, in on_game_loaded
    model.setOpening(redetermine=True)
  File "/home/tamas/pychess/lib/pychess/Utils/GameModel.py", line 360, in setOpening
    opening = get_eco(self.getBoardAtPly(ply).board.hash, exactPosition=True)
  File "/home/tamas/pychess/lib/pychess/Utils/eco.py", line 48, in get_eco
    cur.execute(select, (qh, qhkey, qlang, qpos))
sqlite3.OperationalError: no such column: endline
```

gbtami:
After running pgn2ecodb.sh everything is OK :)

PR #1777, opened 2019-10-01
Fix a few typos in various dialogs/popups

bboutkov:


gbtami:
.pot is generated file. You should make fixes in .py and .glade files instead :)

bboutkov:
Ah always forget which ones are which, thanks. Should be good now.

gbtami:
Great, thx!

Issue #1688, opened 2018-10-08
Replay/Reset one lesson


ecrucru:
Hello,

Once a lesson is finished, it cannot be replayed.

Is it possible to reset an individual progression (instead of everything), or allow to replay a completed lesson ?

Thanks

gbtami:
No reset for individual progression was implemented yet, but it would be useful.
You can replay completed lessons if you open Learn perspective and double click on a selected lesson row.

ecrucru:
Oh sorry, initially, I wanted to say "puzzle" instead of "lesson" :

![image](https://user-images.githubusercontent.com/24614488/46765354-152f6100-ccdf-11e8-8bf1-3b33d9819f67.png)

I finally found that error :
```
Traceback (most recent call last):
  File "/pychess/lib/pychess/perspectives/learn/PuzzlesPanel.py", line 115, in row_activated
    start_puzzle_from(filename)
  File "/pychess/lib/pychess/perspectives/learn/PuzzlesPanel.py", line 131, in start_puzzle_from
    index = progress.index(0)
ValueError: 0 is not in list
```

LessonPanel is written like PuzzlePanel, so the issue is there also :

```
Traceback (most recent call last):
  File "/pychess/lib/pychess/perspectives/learn/LessonsPanel.py", line 100, in row_activated
    start_lesson_from(filename)
  File "/pychess/lib/pychess/perspectives/learn/LessonsPanel.py", line 112, in start_lesson_from
    index = progress.index(0)
ValueError: 0 is not in list
```

Thanks

gbtami:
I think we should add a new "Reset button" to each row next to the progress bars and rename the existing all in one to "Reset all my progress" or something similar.

gbtami:
Fixed above errors in https://github.com/pychess/pychess/commit/e45a6cb9b6b88402a1e6f0c71671039b6446b8db

mathieucaroff:
Hello, I'm a student (with [@Bracewind](https://github.com/Bracewind)). I'm interested in adding this "Reset one lesson" feature.

### Feature definition

@gbtami said:

> I think we should add a new "Reset button" to each row next to the progress bars and rename the existing all in one to "Reset all my progress" or something similar.

I agree. I think it might be nicer to have a "Show per-lesson reset buttons" checkbox. It would be unticked by default, so that the individual reset buttons are not visible by default. That checkbox's state would be temporary, lost when the application is closed, possibly lost when leaving the lesson view.

[...]

PR #1793, opened 2019-11-12
Allow resetting a single puzzle row

mathieucaroff:
Targets issue #1688.

* Add a column of CellRendererPixbuf to the PuzzlePanel.
* Change the function row_activated to let it handle the reset buttons.
* When the button is clicked, the new method _reset_progress_file queries the user and handles resetting a single progress file.

Here's what it looks like:
PyChess puzzle panel screencap

Note: The current change targets solely the puzzle panel. The lesson panel will be addressed later.


gbtami:
Great, thx!

PR #1794, opened 2019-11-13
Make the PuzzlesPanel Reset column the rightmost

mathieucaroff:
Targets issue #1688.

In PuzzlesPanel, move the Reset column to the right, now that I know how to [set which column gets expanded](https://discourse.gnome.org/t/treeview-column-auto-sizing-behaviour/2112).


gbtami:
Thx!

PR #1796, opened 2019-11-13
Disable reset in PuzzlesPanel when progress is 0

mathieucaroff:
Targets issue #1688.

To do this, I added a field to the TreeView store. The new field specifies whether or not the icon should be shown. The `_reset_progress_file` method also checks whether the icon exists before prompting the user. I had to extend the code which computes the progress, but it was duplicated, so factorized the logic into a method: `_compute_progress_info`.

This PR also does the following:

- The `stock_id` property of Pixbuf was actually deprecated so I switched to the new `icon_name` property and created the `GTK_ICON_VIEW_REFRESH` constant to hold the icon name. See (https://lazka.github.io/pgi-docs/Gtk-3.0/constants.html#Gtk.STOCK_REFRESH)
- The internationalization of the modal message did not use interpolation as it should. This is fixed in this PR.

### DIFF
--- a/lib/pychess/Utils/const.py
+++ b/lib/pychess/Utils/const.py
@@ -231,5 +231,8 @@ MENU_ITEMS = GAME_MENU_ITEMS + ACTION_MENU_ITEMS + VIEW_MENU_ITEMS + EDIT_MENU_I
 # Column name
 COLUMN_ROW_RESET = "column_row_reset"
 
+# Gtk icon name
+GTK_ICON_VIEW_REFRESH = "view-refresh"
+
 # Learn categories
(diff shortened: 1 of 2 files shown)


gbtami:
Thx!

gbtami:
There is something wrong with this patch.

```
  File "/home/tamas/pychess/lib/pychess/perspectives/learn/PuzzlesPanel.py", line 121, in row_activated
    self._reset_progress_file(filename, title)
  File "/home/tamas/pychess/lib/pychess/perspectives/learn/PuzzlesPanel.py", line 141, in _reset_progress_file
    _('This will reset the progress to 0 for the puzzle "{title}"').format(title))
KeyError: 'title'
```

PR #1798, opened 2019-11-14
Modern CPU

ecrucru:
Hello,

This PR proposes an easy fix for checking under Windows if the cpu is modern.
The check is imperfect, but it is less wrong than the current one.
You should now be proposed "stockfish_popcnt.exe" in "add engines in mass" if you have a 64-bit cpu.

HaGN

### DIFF
--- a/lib/pychess/System/cpu.py
+++ b/lib/pychess/System/cpu.py
@@ -24,7 +24,8 @@ def get_cpu():
         result['popcnt'] = 'popcnt' in info
         result['bmi2'] = 'bmi2' in info
     except OSError:
-        # TODO Windows
-        result['popcnt'] = False
-        result['bmi2'] = False
+        # Logic not fully true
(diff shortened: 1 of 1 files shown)


gbtami (changes requested):
I think assuming bmi2 is too much. Even my own ThinkPad T510 laptop has no bmi2 support :)
But I can agree that we can assume popcnt support nowadays.

ecrucru:
BMI2 is ~5 years old. In your situation, you should exclude manually few proposed engines while the current version proposes nothing at all. The purpose is just listing the most relevant engines and few of them use this instruction set. Please expect no further analysis from me on that tiny topic. HaGW

gbtami:
I was hallucinating that this PR will lead to incorrect downloads of scoutfish and chess-db. I'v just realized now that they doesn't have bmi2 binaries at all. Sorry for the misunderstanding!

PR #1801, opened 2019-11-27
Updated remote games

ecrucru:
Hello,

This is a technical update for the remote games to cover more links and some special moves.

HaGN


gbtami:
Thx!

Issue #1686, opened 2018-10-08
Show the threatened/unprotected pieces on the board


ecrucru:
Hello,

Highlighting the threatened pieces (attacked and not protected) and the unprotected pieces (no attack and no defense) on the board could be a way to learn passively. Good players generally succeed in protecting _all_ their pieces.

This would be an option. Online, the feature is to be shut down of course.

Thanks

Bracewind:
I'm trying to implement this issue, and for that I want to use getAttacks in Utils.lutils.attacks.py. 
From what I've understood, it takes a lboard, a cord represented as a number and a color.
But it returns a bitboard, and I don't really get how the board is represented using this object...
There is no doc in bitboard.py about the meaning of this object, could you explain me what is it and how it works ?

gbtami:
https://www.chessprogramming.org/Bitboards

Bracewind:
This is my answer to Ecrucru, see the pull request #1803 for the reference

1. This is why I hate implicit polymorphism, I added the support algorithm to SetupModel. Don't know if I need to add it in TimeModel. It would be great if there was an interface to know exactly which functions and attributes each model need to implement to use it in BoardView. After adding this line, it works fine.

2. 
Just playing against the AI to test the new feature. Let's consider the position : r3k2r/ppp2ppp/4p3/3PPb2/8/2P1b1P1/PP4BP/2KR2N1 w kq - 3 17. I play Rd2 and premove Kxd2. AI plays Bxg1 and the moved bishop to g1 is not redrawn on the board.

There is a bug there, because the attribute shown in gameModel class is not what I supposed it was, aka an integer corresponding to the move we want to display. But its value is 33 while there is only one move that has been played. **Unless I know exactly what shown is,** @gbtami  I can't solve this bug.

3.
I use pyChess to create studies. When I transcribe paper games to computer, I play "human vs human" and I would appreciate your new feature to better see the threats that may be relevant for the PGN notation. Imho, the new feature should be disabled if it is not activated or if we are playing online.

Not sure if I understand what you want. But I modified the code so that all local games use the algorithm when it is activated.

4
The checkbox added in the preferences should have its caption beginning with a letter in uppercase. That's consistent for the UI and the preparation of the translations.

Done, 

I will do a Pull request after the points 2 and 3 become clearer.

ecrucru:
Hello,

1. OK

[...]

PR #1803, opened 2019-12-10
Add threatened and unprotected pieces

Bracewind:
Target : #1686

Details of the implementation:
This PR includes those functionnalities :

- Finding the pieces that are in danger thanks to the class DecisionSupportAlgorithm and functions in attacks.py file
- Activating the option only against IA thanks to the GameModel class
- Displaying the different coordinates in danger using drawSupportAlgorithm in BoardView class.
The function is called in draw method of the BoardView class.

The class DecisionSupportAlgorithm is called only once per turn. The gameModel class is responsible of indicating to the first class that the turn ended.
The function drawSupportAlgorithm uses circles to show to the user the pieces in danger. It uses the circle drawing function already implemented. Yellow circles are used to indicate unprotected pieces and red ones are used for threatened pieces.

After implementing it, I found out that it shows the towers at the beginning of the game, because they are not protected by any pieces. This behaviour can be changed in DecisionSupportAlgorithm if needed.

Next step of the PR : allow users to choose when to activate the functionnality


gbtami:
I find two minor issues.
1. While I move in game/move history circles are not updated.
2. Attacked pieces not always indicated with red circle.
For example after 1.e4 e5 2.Bc4 I expected red circle on f7.
It is not clear for me that red circles are for hanging attacked pieces or for anything attacked.

Bracewind:
The algorithm works so that it shows the attacked pieces of the current player, not the enemy pieces. 
Moreover, if the piece is protected but a piece with strictly less value attack this piece, the piece is still considered as being in danger.

For the history, I can't find where we choose the right moves in the moves array...

I have implemented the choice in preferences for enabling/disactivating the algorithm

gbtami:
The preferences option not activated on pychess next start. You can find how several other config options are get/set in https://github.com/pychess/pychess/blob/master/lib/pychess/widgets/BoardView.py

Seems there are no red circles at all now.

setShownBoard() is called from several panels when you move away from current position.
https://github.com/pychess/pychess/blob/master/lib/pychess/widgets/BoardView.py#L490

Bracewind:
Normally, I've added the red circles, and the preferences are taken into account at the start of the game. 
Moreover, the algorithm is run again when going back in history.

gbtami:
Just some minor things I noticed.
Instead of `game_against_ia` you can use `hasEnginePlayer()` in Gamemodel.py
Instead of `CORD_CONST` you can use `cordDic` from const.py

I suggest not to draw circles on Kings at all. They can't be captured in standard chess anyhow :)

I noticed one freeze while testing your PR. Not the whole GUI was frozen, just I was unable to make my next move. Unfortunately I can't reproduce this again. (I can imagine it is not related to your changes at all.)

Bracewind:
For the freezing stuff, it could be linked. When I coded the functionality, the Gui froze when I messed up. It is because redraw call my draw function, which call redraw etc..... It works fine as long as new coordinates are not introduced before the end of the first call of my drawing function.  It should probably be modified to not call redrawCanvas, but some coordinates are not redrawn if I don't explicitly call it.... And when I want to apply the algorithm on a move already done using the history, because GameControl is not called back when changing the move shown, I must call the algorithm in Boardview. 


I modified the code according to your advice

gbtami:
I think we can avoid freeze issues with a little conceptional change. `drawSupportAlgorithm()` shouldn't be a **drawing** function at all! It is enough to put all Cords returned by `algorithm.calculate_coordinate_in_danger()` into BoardView circles set.
I mean instead of drawing anything just do 
```
        for cord in coord_attacked_not_protected:
            self.circles.add(cord)
``` 
And instead of calling  `drawSupportAlgorithm()` inside `draw()`
just call it inside `_setShown()` something like:
```
        if self.drawSupportAlgorithm(shown):
            need_redraw = True
```
before
```
        if need_redraw:
            self.redrawCanvas()
```

Bracewind:
The array containing the circles is cleared each time we click on the right button of the mouse, which is not a behaviour we want for visualizing the result of the algorithm.

I've done something very similar to what you said : 
- in set shown, the algorithm is applied and need_redraw is called
- in draw function, there is still a draw function for the algorithm, but it never call redraw_canvas and retrieve the coordinate directly from the algorithm. Then it draw circles at the given coordinates.

gbtami:
I think this is OK now. Great work! Thx a lot for this contribution!

ecrucru:
Hello,

I tested the new feature and it is impressive, that's very nice. And I would like to share few comments with you.

1) Click "Game > New game > From custom position" :

```
  File "/pychess/lib/pychess/Main.py", line 267, in on_set_up_position_activate
    newGameDialog.SetupPositionExtension.run(fen, variant)
  File "/pychess/lib/pychess/widgets/newGameDialog.py", line 757, in run
    setup_position=True)
  File "/pychess/lib/pychess/widgets/BoardControl.py", line 41, in __init__
    self.view = BoardView(gamemodel, setup_position=setup_position)
  File "/pychess/lib/pychess/widgets/BoardView.py", line 244, in __init__
    self.onSupportAlgorithmActivation()
  File "/pychess/lib/pychess/widgets/BoardView.py", line 484, in onSupportAlgorithmActivation
    self.model.support_algorithm.enableDisableAlgo(activation_wanted)
AttributeError: 'SetupModel' object has no attribute 'support_algorithm'
```

2) Just playing against the AI to test the new feature. Let's consider the position : `r3k2r/ppp2ppp/4p3/3PPb2/8/2P1b1P1/PP4BP/2KR2N1 w kq - 3 17`. I play Rd2 and premove Kxd2. AI plays Bxg1 and the moved bishop to g1 is not redrawn on the board.

3) I use pyChess to create studies. When I transcribe paper games to computer, I play "human vs human" and I would appreciate your new feature to better see the threats that may be relevant for the PGN notation. Imho, the new feature should be disabled if it is not activated or if we are playing online.

4) The checkbox added in the preferences should have its caption beginning with a letter in uppercase. That's consistent for the UI and the preparation of the translations.

Thanks in advance for any further changes on these points.

HaGN

PR #1818, opened 2020-04-11
Fix all DeprecationWarning: invalid escape sequence

BoboTiG:
Hello,
This is a clean-up patch fix all deperaction warnings: invalid escape sequence. I used that command to catch them all:
```shell
PYTHONWARNINGS=all python -m compileall -f -q *
```


gbtami:
Great, thx a lot!

Issue #1811, opened 2020-03-05
AttributeError: 'TreeModelFilter' object has no attribute 'sort_new_with_model'


ghambo:
Manjaro, Kde, imho all dependencies are installed (it was working, but suddenly stoped...)

 

![Screenshot_20200306_003613](https://user-images.githubusercontent.com/7389117/76024047-dc330480-5f43-11ea-926d-6fb1c1729fd3.jpg)
![Screenshot_20200306_144438](https://user-images.githubusercontent.com/7389117/76077007-59e52780-5fb9-11ea-8b1b-ea2a4e2d5d8e.png)

gbtami:
Suspicious traceback line:
AttributeError: 'TreeModelFilter' object has no attribute 'sort_new_with_model'

This seems not FICS connection related issue, but some Gtk incompatible change? (Grrr)

What are your Gtk etc. versions on second tab here?
![http://i.imgur.com/NEslG3R.png](http://i.imgur.com/NEslG3R.png)

ghambo:
![Screenshot_20200306_144438](https://user-images.githubusercontent.com/7389117/76077007-59e52780-5fb9-11ea-8b1b-ea2a4e2d5d8e.png)

ghambo:
imho its gtk problem...
Tanks in advance

BoboTiG:
This is a bug in [GTK](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/1134) that is fixed in the `pygobject` module with that [commit](https://gitlab.gnome.org/GNOME/pygobject/-/commit/0b1ff4d94cd811f6fbf8ab9f23ff4b2e6e80897a). Unfortunately this will not be available right now as the next `pygobject` release is not yet planned.

I will work on a workaround inside PyChess directly.

BoboTiG:
@ghambo could you try my patch (#1822) and let us know if that works for you? I tried on Manjaro 19.0.2 KDE and reproduced the bug without the patch; and with the patch all seems to work fine. But I would like your confirmation first :)

mirao:
Thanks. I can confirm the patch works for me (I reported the same issue in #1815 and just tested that the issue is still present without the patch).

Verified with
- Ubuntu 20.04
- gnome-shell 3.36.1-5ubuntu1
- python3-chess 1.0.0-1 + patched file `/usr/lib/python3/dist-packages/pychess/perspectives/fics/__init__.py`

ghambo:
Great job!
[...]

PR #1820, opened 2020-04-12
Remove use of deprecated configparser.readfp() for read_file()

BoboTiG:
Hello,

Should be the last clean-up patch :)
[readfp()](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.readfp) is deprecated since Python 3.2.

### DIFF
--- a/lib/pychess/System/conf.py
+++ b/lib/pychess/System/conf.py
@@ -24,7 +24,7 @@ path = addUserConfigPrefix("config")
 encoding = locale.getpreferredencoding()
 if os.path.isfile(path):
     with open(path, encoding=encoding) as fh:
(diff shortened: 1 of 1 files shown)


gbtami:
Thx!
Btw if you are interested to fix more issues in pychess may I ask you to take a look on https://github.com/pychess/pychess/issues/1811 please?
Telling the truth I spent most of my free time in last year developing chess variant server https://github.com/gbtami/pychess-variants but pychess needs a new release also and that is a real blocker.
Thx again for all of your fixes so far!

BoboTiG:
Yeah of course, I will have a try when I can ;)

PR #1821, opened 2020-04-13
Fix more DeprecationWarning: unclosed file

BoboTiG:
I missed some warnings, so here is another patch for unclosed file warnigs.


gbtami:
Thx!

PR #1824, opened 2020-04-25
/usr/share/appdata is deprecated in favor of /usr/share/metainfo

chewi:
See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent.


gbtami:
Thx!

PR #1825, opened 2020-04-25
Don't pre-compress the man page

chewi:
Having a pre-compressed file is not good for source control and is
unhelpful for distributions, who may want to apply their own
compression scheme. This is only saving about 300 bytes anyway.


gbtami:
Thx!

gbtami:
While I agree with you the goals I had to revert this commit because it broke rpm creation with
setup.py bdist_rpm

File not found: /home/../pychess-master/build/bdist.linux-x86_64/rpm/BUILDROOT/PyChess-1.0.1-1.py38.x86_64/usr/share/man/man1/pychess.1

If you can fix create_rpm_py38.sh feel free to send another PR.

chewi:
Gah, this is due to pypa/setuptools#1277, which stems from this [18 year old bug](https://bugs.python.org/issue644744). I'll try to think of a workaround but I don't think there is one.

gbtami:
OMG :)

rmathar:
There are RPM distributions (CentOS, openSUSE) which have a macro `ext_man` defined. So I tend to insert

```
%if 0%{?fedora_version}
%{!?ext_man: %define ext_man .gz}
%endif
```
before the `prep` in the PRM spec-file and to use a scheme like
```
%{_mandir}/man1/%{name}.1%{?ext_man}
```
in the `%files` section of the spec-file. One would have to find a way to derive from that value of the `ext_man` which compressor is to be applied to generate the compressed man-page that is supported by the various OS flavor.s

PR #1839, opened 2020-06-28
Fix Websockets

ecrucru:
Hello

By switching to async code, all WS-based chess providers should now work in the foreground, but no more in the background.

It also needs Python 3.5 else the WS are just turned off.

HaGN


gbtami:
Thx, great fix!
If you are intereseted about using async await, see #1840 and #1841 
We definitely should drop 3.4 support right now!

PR #1849, opened 2020-07-04
Charge of the light brigade

ecrucru:
Hello,

This PR implements the chess variant [Charge of the light brigade](https://www.chessvariants.com/rules/charge-of-the-light-brigade) that is playable with FairyMax.

It uses the normal chess rules with a special setup, but White promotes to Q and Black to N.

Happy week-end


gbtami:
Awesome! Thx a lot!

PR #1865, opened 2020-08-09
Technical update

ecrucru:
Hello

This is a tiny update for:

- the remote games: new url and ability to call 2 popular sites faster from the single id of the games
- a translation: a parenthesis was not set at the right location

Happy weekend


gbtami:
Thx!

PR #1894, opened 2021-04-23
Update pychess.po

ghost:
little French mistake

### DIFF
--- a/lang/fr/LC_MESSAGES/pychess.po
+++ b/lang/fr/LC_MESSAGES/pychess.po
@@ -3265,7 +3265,7 @@ msgstr "Parce que %(loser)s a dépassé son temps réglementaire"
 #: lib/pychess/Utils/repr.py:60
 #, python-format
 msgid "Because %(loser)s was checkmated"
-msgstr "Parce %(loser)s était échec et mat"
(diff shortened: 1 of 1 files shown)


gbtami:
Pychess uses Transifex for translations. Go to https://www.transifex.com/projects/p/pychess/ please! Thx.

sshivaji:
Laripette is right in this case as checkmate is a sudden and connected
action and not something that was there for a long time.

Is the solution to go to transifex and fix these strings?

Thanks!

gbtami:
Yes. See https://github.com/pychess/pychess/wiki/TransifexTranslates

Transifex is the source of our translations. Once per day transifex auto merges from https://github.com/pychess/pychess/blob/master/lang/pychess.pot
Translators work on it on transifex web GUI, and before releases (or occasionally) I pull new translations from there to our github.

Here is a screenshot how it looks like on transifex

sshivaji:
Thanks, will follow up, ideally, Laripette can fix it on Transifex. Even
though French is not my native language, Parce and était are quite annoying
to read! :)

ghost:
I don't have a transifex account, if someone could correct the mistake for me.

thank you

PR #1902, opened 2021-05-19
Allow linux executables in new engine dialog

### LINKED ISSUE #1901: Unable to select a linux executable in the new engine dialog
Hi,

I'm on linux and I was trying to add stockfish to pychess and the new engine dialog doesn't allow you to add linux executables. The file picker dialog filters them out:

Screenshot from 2021-05-19 11-34-35

It's because the right mimetype `x-sharedlib` isn't in the filter list. Adding it should fix the issue like so:

Screenshot from 2021-05-19 11-35-54

rbobrowicz:
Fixes #1901 

Adds the mimetype for linux executables to allow you to add new chess engines on linux.

Also adds a second file filter for "All Files" (which is usually good HIG guidelines anyway) in case users have executables with some other mimetypes as well.


rbobrowicz:
Wait, I'm a moron I based my patch on an outdated version of pychess that already has this fixed.

PR #1908, opened 2021-07-31
Remove the nonexsiting 'Pychess' theme from themes. Fixes #1907

### LINKED ISSUE #1907: discoverThemes() in widgets/preferencesDialog.py adds a nonexitent "Pychess" theme to return list
[The function](https://github.com/pychess/pychess/blob/4cef7d80c5df444e143c8ba7ba7db6098a3eb028/lib/pychess/widgets/preferencesDialog.py#L800) starts with
`    themes = ['Pychess']`
and then continues to discover themes in the pieces directory and append them all to themes list. This "Pychess" theme does not exist, and as one single invalid theme causes the for loop in [here](https://github.com/pychess/pychess/blob/4cef7d80c5df444e143c8ba7ba7db6098a3eb028/lib/pychess/widgets/preferencesDialog.py#L752) to abandon showing the rest of the list (because of that break), any theme that is after "Pychess" alphabetically will not be shown in the themes tab in preferences.

masoudd:
The discoverThemes() adds a 'Pychess' theme by defualt this causes the loop that displays the themes in Preferences->themes tab to terminate prematurely and leave out any theme that comes after 'Pychess' alphabetically.
See issue #1907

### DIFF
--- a/lib/pychess/widgets/preferencesDialog.py
+++ b/lib/pychess/widgets/preferencesDialog.py
@@ -797,7 +797,7 @@ class ThemeTab:
 
             :return: (a List) of themes
         """
-        themes = ['Pychess']
+        themes = []
 
         pieces = addDataPrefix("pieces")
         themes += [d.capitalize()


gbtami:
Nice catch, thx! :)

Issue #1899, opened 2021-05-09
pièces chess


ghost:
Hello

It's been several years since the themes have been updated.
Would it be possible to put new themes of better quality?

I have some examples here on the site of lichess:
https://github.com/ornicar/lila/tree/master/public/piece

gbtami:
Pull requests are welcome! :)

ghost:
I have to create a request in the pull requests section ?

ghost:
thank you for the new pieces

gbtami:
Courtesy of @masoudd :)

PR #1909, opened 2021-07-31
Add all lichess pieces that are not in pychess

masoudd:
I added one theme per commit. All svg files are from [lila](https://github.com/ornicar/lila/tree/master/public/piece).
I went through [lila's COPYING.md](https://github.com/ornicar/lila/blob/master/COPYING.md) and added a license.md file to each theme directory (except "riohacha" as I couldn't find any license for it).
Some of the pieces on lila were also in pychess already. I didn't change the existing ones with the lila version.

related issue  #1899


gbtami:
Thx!  Seems all theme is OK except the Horsey one. Can you double check it, please?

gbtami:
They are OK in inkscape and in Firefox when I open them in github.
Seems my Ubuntu 18.04 has some old versions of affected libs.

Issue #1918, opened 2021-09-08
PyChess keeps UCI engine running even if GUI is closed


Javran:
Reproducible in 1.0.3, linux, python 3.9.7:

- config a UCI engine (stockfish should do)
- start a game with that engine
- close PyChess

Both PyChess and stockfish processes keep running despite closing.
There are error messages from stderr:
```py3
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/pychess/Main.py", line 371, in on_quit1_activate
    if perspective.closeAllGames(perspective.gamewidgets) in (
  File "/usr/lib/python3.9/site-packages/pychess/perspectives/games/__init__.py", line 440, in closeAllGames
    game.terminate()
  File "/usr/lib/python3.9/site-packages/pychess/Utils/GameModel.py", line 984, in terminate
    player.end(self.status, self.reason)
  File "/usr/lib/python3.9/site-packages/pychess/Players/UCIEngine.py", line 124, in end
    self.parse_line_task.cancel()
TypeError: cancel() takes 1 positional argument but 2 were given
```
What I suspect is the `parse_line_task.cancel()` is called with incorrect arguments and the coroutine is never terminated, so does the main event loop. I know a bit of Python but never worked with asyncio close enough to understand how on earth is `cancel()` called with no arguments but it complained "2 were given" (one is the object another is `msg`?), but I hope this information helps troubleshoting the problem.

masoudd:
@Javran Can you test if this is fixed by this commit?

gbtami:
Should be OK now. If this is not the case feel free to reopen.

Javran:
@masoudd Thanks for the fix! I can confirm the error is gone now.

PR #1929, opened 2022-01-08
Add the new optional `msg` parameter to `cancel` method of the class pychess.Util.wait_signal

masoudd:
Since python 3.9, the `cancel` method of Futures has
an added optional `msg=None` parameter:
https://docs.python.org/3/library/asyncio-future.html#asyncio.Future.cancel

See issue #1918
After adding this optional parameter, the `TypeError` goes away.


gbtami:
Thx!

PR #1932, opened 2022-02-02
Remove all files in pieces/ttf/*.ttf

### LINKED ISSUE #1930: double pieces
in the last version, some parts are duplicated.

image

masoudd:
if the pieces/ttf directory itself is deleted then
the create_theme_preview.py fails.
I can edit that script to fully delete all traces of the ttf directory if needed.

Closes #1930


masoudd:
There is also some ttf and cairo related logic in `lib/pychess/gfx/Pieces.py` like [get_chess_font_face(name)](https://github.com/pychess/pychess/blob/e3f9b12947e429993edcc2b9288f79bc181e6500/lib/pychess/gfx/Pieces.py#L124) and

[if piece_set.startswith("ttf-"):](https://github.com/pychess/pychess/blob/e3f9b12947e429993edcc2b9288f79bc181e6500/lib/pychess/gfx/Pieces.py#L160)

Should I rip it all out or just leave it be?

gbtami:
Yea, we should remove all ttf related code as well I think.

masoudd:
removed around 200 lines. I am unsure about the changes in `setup.py`. Please check if it's ok

gbtami:
Thx!

Issue #1863, opened 2020-07-30
Add "Skip puzzle" putton


gbtami:
Splitted from https://github.com/pychess/pychess/issues/1862
Now, when the engine gives a false hint to a puzzle and even if i make the correct move it says retry. That means i can't move to the next puzzle. Perhaps can you consider adding a skip puzzle option? What i am at present doing is keep going back to the .olv file keep deleting the incorrect hints ones and then save it and then restart the application and play from the next puzzle. That also looks a bit tedious.

Arunachala-Ramana:
Great. skip puzzle + previous puzzle button would be great, so that one can go and revise any puzzle he wants. This should be there in any instructional lessons, be it endgames tactics or puzzles.

gbtami:
Fixed via https://github.com/pychess/pychess/pull/1946

PR #1938, opened 2022-04-24
Skip puzzle or lesson

jonakeys:
In order to address the request for enhancement https://github.com/pychess/pychess/issues/1863 I tried to solve it.


gbtami:
First of all sry for the late reply.
I think the "Skip" button should be a blue text button after "Best move" (or before "Hint") here, and also should call the "Next button" to open the next puzzle

jonakeys:
@gbtami Thanks for your reply and it's no problem it took a while. I removed the previous additions and I'm now going to resolve it using the directions you gave.

Issue #1934, opened 2022-02-04
Choose depth of analysis


ghost:
it would be nice if we could set the depth.
Here we can only set how long it will analyze.
![image](https://user-images.githubusercontent.com/77933011/152600150-52b8781f-6f95-4d6a-9dcf-345ef6bb5b13.png)

TheRedLancer:
I think I can make this happen!

gbtami:
Fixed via https://github.com/pychess/pychess/pull/1939

PR #1939, opened 2022-04-27
Add preference to limit analysis by depth

TheRedLancer:
I added a SpinButton for choosing the depth of engine analysis, and a CheckBox for toggling. I added a couple more config values for the depth limit value and whether to limit by depth. I added code for checking that value to the same places the code already checked for time-limited analysis. I changed how the UCIEngine handles a finite analysis time from sending a "go movetime" command to sending a "stop" command after an asyncio.Loop finishes. The CECPEngine already used an asyncio.Loop for stopping the analysis so I figured the UCIEngine could work the same.

Addressing issue: #1934


gbtami:
Great, thx!

Issue #1940, opened 2022-04-27
analyze


ghost:
thanks for the update.
when you enter a 3 digit number, there is a visibility problem.
![Capture d’écran du 2022-04-27 14-41-25](https://user-images.githubusercontent.com/77933011/165520466-c6b6babb-5954-4569-b021-a4b185f02047.png)

how to know which option is activated ?
Is it the time of the analysis or the depth ?

TheRedLancer:
I can fix the UI spacing, I never use a depth higher than ~40. The 2 options work independently. If you have it set for a depth of 20 and a time limit of 30 seconds, whichever limit the system hits first will stop the analysis. If the engine gets to depth 20 in 25 seconds, the analysis will stop. If it only gets to 18 in 30 seconds the analysis will stop as well.

gbtami:
Limiting the spin to 99 max seems enough for me.

ghost:
can we choose one or the other.
If for example I just want to choose the depth and not the duration of the analysis.

a selection with two buttons.
![Capture d’écran du 2022-04-27 23-14-49](https://user-images.githubusercontent.com/77933011/165632565-536d3233-80b8-49f1-ad7f-ab4cb1cc80d9.png)

TheRedLancer:
You can choose one or the other. If you want to limit by time, you can select infinite depth. If you want to limit by depth, you can select the other check box. I thought about using the Radio buttons, but I reasoned that this implementation of using 2 checkboxes allowed for more flexibility.

PR #1941, opened 2022-04-27
Limit analysis depth to 99

TheRedLancer:
I limited the Glade element to only allow for 2 digits, thus limiting the analysis depth to 99.

Issue #1940


gbtami:
Thx!

PR #1955, opened 2022-07-17
Drop legacy bits: Debian menu, SVG icon in `/usr/share/pixmaps`

pinotree:
Debian officially deprecated the Debian menu system 7 years ago [1] in favour of the freedesktop.org menu.

Hence, drop the old Debian menu file, as there is already a desktop file
that will be enough for (almost) all the Linux users. Drop its XPM icon,
which was never used for it (leaving the menu file with no icon) because
of the wrong filename (`pychess.xmp` vs `pychess.xpm`).

[1] https://lists.debian.org/debian-devel-announce/2015/09/msg00000.html

Also, stop installing `pychess.svg` to legacy pixmaps location: it is already installed in the global hicolor XDG icon theme, so there is no need to install it also in the legacy pixmaps location (and it is not a pixmap, even).


gbtami:
Thx!

PR #1963, opened 2023-03-08
Fix pieces theme preview regressions introduced by #1960

hartwork:
Hi!

There are two related issues addressed here:
- The two pieces theme preview images added in #1960 are not in sync with what `create_theme_preview.py` is rendering, so running `PYTHONPATH=lib python3 create_theme_preview.py` would produce a diff. That's now fixed and the new images also look a bit cleaner regarding the edges and more consistent with the others, as a side effect.
- `create_theme_preview.py` was broken by #1960 for case sensitive file systems (e.g. Linux but not Windows) and hence you would get a warning "Can't create piece set atopdown" and then end up with an image that doesn't match the theme. That's fixed now, easiest way to fix was to just rename the folder.

What do you think?

CC @Xyvir @gbtami 

~PS: We'll need some version of this for a true chance at a green CI from pull request #1964.~


gbtami:
Thx!

PR #1976, opened 2023-03-09
requirements(deps): bump importlib-metadata from 1.6.0 to 6.0.0 + pin new indirect dependencies (alternative to #1966)

hartwork:
Going further with #1966 to get dependencies back to fully pinned and a green CI.


gbtami:
Thx!

PR #1983, opened 2023-03-11
Improve default looks

hartwork:
Hi!

I'm aware that taste is subjective but the current default look is so out of place for 2023 compared to e.g. Lichess defaults that I feel that this minimum set of changes would greatly help the first visual impression that users get of PyChess:

1. Do not draw the grid by default
2. Switch default pieces theme from `Chessicons` to `Cburnett`

What do you think?


gbtami:
Agreed. Thx.

PR #1984, opened 2023-03-11
 requirements(deps): bump pytest from 5.4.2 to 7.2.2 + pin new indirect dependencies (alternative to #1978)

hartwork:
Alternative to #1978

This builds on top of #1978 but also pins the new indirect dependencies that were proven missing at https://github.com/pychess/pychess/actions/runs/4384798067/jobs/7676704521#step:5:182 .


gbtami:
Thx.

PR #1985, opened 2023-03-11
Make PerspectiveManager.on_persp_toggled more robust

hartwork:
Attribute `current_perspective` can be `None` before it is set a few lines below, for the first time.

PS: Can be seen triggered by current CI e.g. at https://github.com/pychess/pychess/actions/runs/4379761622/jobs/7666065791#step:6:69 .


gbtami:
Thx!

PR #2001, opened 2023-03-17
requirements.txt: Drop by-now-unused and vulnerable py==1.11.0

hartwork:
Inspired by https://github.com/pychess/pychess/security/dependabot/3


gbtami:
Thx. I seen this as well, but was not able to remember where we use it :)

PR #2023, opened 2023-03-21
Make GitHub Actions build Linux packages

hartwork:


.github/workflows/run-tests.yml
gbtami, line 1: I think we should split off .deb and .rpm creation from run-tests.yml into two separate workflow, deb-build.yml and rpm-build.yml respectively. The reason is the different need of Python3 versions. While .deb creation needs only one Python version (it can be the ubuntu-22.04 default python3.10), all the Python3 version specific .rpm creation needs all the major Python3 versions from 3.7 to 3.11 (it creates 3.7 and 3.11 packages only atm).

hartwork: @gbtami maybe you're right. We'll waste some resource for runtime but for some clear upsides. I'll start working on that.

For the RPMs, I wonder why we need different packages for 3.7 and 3.11. Is it about the Python interpreter command only? Would it go away if we said `/usr/bin/pyhon3` there? I'm just trying to understand why it even needs more than a single RPM package for Python code.

hartwork: > @gbtami maybe you're right. [..] I'll start working on that.

Done.

gbtami: Contrary to deb, bdist_rpm created rpm contains files in python major version dependent directory structure and also contains .pyc files like this:

hartwork: @gbtami interesting! :+1:

gbtami: http://rpmfind.net/linux/RPM/fedora/devel/rawhide/x86_64/p/pychess-1.0.3-5.fc38.noarch.html

hartwork: > http://rpmfind.net/linux/RPM/fedora/devel/rawhide/x86_64/p/pychess-1.0.3-5.fc38.noarch.html

@gbtami is the link intended to demonstrate that Fedora's own package contains (Python 3.11) `.pyc` files as well? Something else?

gbtami: It is just shows that it contains the Python major version dependent .pyc files as well, yes.

gbtami: .deb seems OK, I tested it.
The generated .rpm files now put everything under /opt/hostedtoolcache/Python/3.10.10/x64
Contrary to the old releases and the linked fedora. They put everything under /usr
Maybe it is OK, but I have the feeling it is not. Idk.
Possible this is caused by the `which` usage in .sh files.

hartwork: @gbtami good finding, the which is related but not the problem.

hartwork: @gbtami I should have put it another way: if we take the which away, then there will be no `/usr/bin/python3.XX` except the `/usr/bin/python3.10` that comes with Ubuntu. So which is the cure and the problem. Any ideas how to provide all of  `/usr/bin/python3.7` to `/usr/bin/python3.11` from the same Ubuntu and have the binary believe that it's actual home, easily?

hartwork: @gbtami we could say goodbye to matching .pyc files for everything but Python 3.10. That would mean a theoretical loss in performance to everyone not on Python 3.10.

hartwork: @gbtami we could build python3.xx Debian packages live in the CI based on…

- 3.7: http://security.debian.org/debian-security/pool/updates/main/p/python3.7/python3.7_3.7.3-2+deb10u4.dsc
- 3.8: gone
- 3.9: http://deb.debian.org/debian/pool/main/p/python3.9/python3.9_3.9.2-1.dsc
- 3.10: http://deb.debian.org/debian/pool/main/p/python3.10/python3.10_3.10.10-2.dsc
- 3.11: http://deb.debian.org/debian/pool/main/p/python3.11/python3.11_3.11.2-6.dsc

…but there won't be Python 3.8 this way and we'd probably have to update these links/versions every few days or weeks.

Another approach would be to compile Python upstream tarballs directly. Do you want me to demo that?

gbtami: You can try to install other Python versions in CI from https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa
It works OK form me. (I'm on 18.04 myself.)

hartwork: @gbtami interesting idea, didn't think of deadsnakes at all. Let me try :+1:

hartwork: @gbtami in my experiments it looks like we'll get:
- `/usr/local` from pre-installed Python 3.10
- `/usr` from deadsnakes.

I found that I can use file `~/.pydistutils.cfg` to influence where files go, e.g. with content…

```ìni
[install]
install-base = /usr/FOO
```

…but I'll get an error then…

> error: install-base or install-platbase supplied, but installation scheme is incomplete

…and if I specify all variables, there's a chance that I could get on of the other values wrong, in particular ones, that are not in use right now but could be used in the future. This may be the complete list:

- `--install-base`
- `--install-data`
- `--install-headers`
- `--install-lib`
- `--install-platbase`
- `--install-platlib`
- `--install-purelib`
- `--install-scripts`

I'm thinking we could either focus on the  `~/.pydistutils.cfg` approach and get it to work even with original `actions/setup-python` or we try forcing apt-get to always use deadsnakes even for already-installed 3.10 by influencing which repository wins the race, no matter the version.

What do you think?

hartwork: @gbtami quick update, I found this for `~/.pydistutils.cfg` now:

```ini
[install]
prefix = /usr/FOO
```

hartwork: @gbtami ready for review now

gbtami (approved):

.github/workflows/run-tests.yml
gbtami: Awesome work. Everything seems OK now. Thx!

PR #2024, opened 2023-03-21
pieces: Delete the only two piece previews

hartwork:
We have `create_theme_preview.sh` to generated previews. Idea here is that either we have all previews in Git or none. If they're not in Git, they cannot go out of sync. @gbtami what do you think?


gbtami:
Yes. Generated files never should be stored in git. This two came here by accident (PR from the creator). My fault.

PR #2038, opened 2023-04-13
Fixed issue when user's home directory is in prefix during installation

lRespublica:
This occurs during build under ALT Linux. On the build system, ~ expands to /etc/src. And the package building breaks

### DIFF
--- a/lib/pychess/System/prefix.py
+++ b/lib/pychess/System/prefix.py
@@ -16,7 +16,7 @@ if getattr(sys, "frozen", False):
     _installed = True
 else:
     home_local = os.path.expanduser("~") + "/.local"
-    if sys.prefix in __file__:
+    if sys.prefix in __file__ and not sys.prefix in home_local:
         for sub in (
             "share",
             "games",


hartwork (commented):
Hi @CausaPrincipalis71, cool that you're working on updating PyChess to 1.0.4 in ALT Linux! :+1: 

Maybe this is all clear to @gbtami but at least I am having trouble understanding the issue and the idea for the fix in detail.
`sys.prefix` is probably effectively `/usr` for you since you're on Linux? For me on Gentoo Linux:

```console
# python <(echo 'import sys; print(sys.prefix)')
/usr/lib/python-exec/python3.10/../../..
```

Is the build log with the error online somewhere?

PS: The use of `X in Y` to implement "is path Y a subdir of X" seems rather troublesome to me i.e. it may yield both false positives and false negatives. Maybe not something we need to fix right here, but let's see.

gbtami:
Yea, before you send a PR to fix something not trivial, please open an issue and explain what is the bug/problem you want to fix.

hartwork:
@CausaPrincipalis71 can I do anything to help move this forward? PS: I'm available for voice chat in German or English as needed.

lRespublica:
Oh, sorry I didn't see messages.
I will describe problem in detail tomorrow.  It is related to the build system specific to ALT Linhx

lRespublica:
[pychess-build.log](https://github.com/pychess/pychess/files/11260336/pychess-build.log)
A bit of context: during the building of the rpm package, an isolated environment is created. It has a builder user whose home directory is /usr/src. The package is built in /usr/src/RPM/BUILD.

During the build, prefix.py is called. It finds that prefix.py is located in sys.prefix but pychess itself is not installed. After that it raise an exception. 

The patch avoids raising of exception in such case.

hartwork:
@CausaPrincipalis71 thanks for elaborating!

After some investigation I find that:
- The build fails at `pgn2ecodb.py` which is generating needed file `eco.db`.
- There are two source tarballs for 1.0.4:
  - a) https://github.com/pychess/pychess/archive/refs/tags/1.0.4.tar.gz — generated by GitHub using `git archive`, does not contain generated files
  - b) https://github.com/pychess/pychess/releases/download/1.0.4/pychess-1.0.4.tar.gz — generated by our CI, comes with pre-generated file `eco.db` (and also pieces theme preview images like `pieces/Spatial.png`).
- (On a side note, PyChess 1.0.4 is the first release to not generate these files from within `setup.py` automatically, commit 136ef01b6e89ad3f89a79cf0e7e211df86b3fb2e is related if curious.)
- Plain `pip install -e .` worked without a call to `lib/pychess/System/prefix.py` locally.

So how about this for a fix:
- Using https://github.com/pychess/pychess/releases/download/1.0.4/pychess-1.0.4.tar.gz for a source tarball
- Not calling `pgn2ecodb.py`, not calling `create_theme_preview.py`.
- Keeping `lib/pychess/System/prefix.py` unpatched in ALT packaging.

Would this work? What do you think?

lRespublica:
I don't think it will help much. Because `pychess-1.0.4.tar` is obtained directly from git tag 1.0.4 (building from tag is an ALT policy), so it would be better to leave this mechanism.

I tried to build manually using the .tar.gz generated by your CI. This works fine both with pip and with pyproject build and prefix.py is not called.

But still I would like to find a way to build directly from the tag

hartwork:
> I don't think it will help much. Because `pychess-1.0.4.tar` is obtained directly from git tag 1.0.4 (building from tag is an ALT policy), so it would be better to leave this mechanism.

I found this comment at http://www.sisyphus.ru/en/srpm/Sisyphus/pychess/spec now:

> ```
> # Source-url: https://github.com/pychess/pychess/releases/download/%version/pychess-%version.tar.gz
> ```

That would be the enriched version. What is the benefit of avoiding the enriched source tarball? Do you have a link to a page where I could read more about the details and motivation behind that policy? If it was about executable binaries, I'd get it, I'm [packaging for Gentoo Linux](https://packages.gentoo.org/maintainer/sping@gentoo.org) myself. Especially with things like `make dist` of Automake in mind, I would like to better understand it. 

> I tried to build manually using the .tar.gz generated by your CI. This works fine both with pip and with pyproject build and prefix.py is not called.

Sounds great.

> But still I would like to find a way to build directly from the tag

Using a temp directory could be an option.
Or we could exclude `/usr/src` for a parent in code.

hartwork:
@gbtami I'm considering to fix the is-child-path checks using a helper function like this:
```python
def path_is_child_of(parent_path: str, candidate_child_path: str) -> bool:
    real_parent_path = os.path.realpath(parent_path, strict=False)
    real_candidate_child_path = os.path.realpath(candidate_child_path, strict=False)
    return os.path.commonpath([real_parent_path, real_candidate_child_path]) == real_parent_path
```
What do you think, should I make this a separate pull request?

PS: I'm avoiding [`os.path.normpath`](https://docs.python.org/3/library/os.path.html#os.path.normpath) and [`os.path.abspath`](https://docs.python.org/3/library/os.path.html#os.path.abspath) because of ["This string manipulation may change the meaning of a path that contains symbolic links."](https://docs.python.org/3/library/os.path.html#os.path.normpath) and [`os.path.commonprefix`](https://docs.python.org/3/library/os.path.html#os.path.commonprefix) because it's broken by design.

gbtami:
Sure, a separate PR is OK.

lRespublica:
> Do you have a link to a page where I could read more about the details and motivation behind that policy?

I could not find a single source where all this is described. But in short: the main advantage is the ease of updating and transferring the package to [other](https://www-altlinux-org.translate.goog/Gear/remotes?_x_tr_sl=ru&_x_tr_tl=en&_x_tr_hl=ru&_x_tr_pto=wapp) maintainers. This is not mandatory, but highly recommended.

> Using a temp directory could be an option.

The isolation tool does not allow you to do this directly. And that would be overcomplicating of the process.

> Or we could exclude /usr/src for a parent in code.

This is what was added to the PR. Although, perhaps it should be implemented not so rudely

hartwork:
> Sure, a separate PR is OK.

@gbtami cool, done: #2048.

> > Do you have a link to a page where I could read more about the details and motivation behind that policy?
> 
> I could not find a single source where all this is described. But in short: the main advantage is the ease of updating and transferring the package to [other](https://www-altlinux-org.translate.goog/Gear/remotes?_x_tr_sl=ru&_x_tr_tl=en&_x_tr_hl=ru&_x_tr_pto=wapp) maintainers. This is not mandatory, but highly recommended.

I believe that `https://github.com/pychess/pychess/releases/download/%version/pychess-%version.tar.gz` works well for hand-over to other developers, personally.

> > Or we could exclude /usr/src for a parent in code.
> 
> This is what was added to the PR. Although, perhaps it should be implemented not so rudely

Let me improve the base code first then, and we can revisit this case here after #2048.

hartwork:
@CausaPrincipalis71 with #2048 merged now, addition of `and not _path_is_child_of('/usr/src', __file__)` would be an option in theory.

hartwork:
@CausaPrincipalis71 what do you think?

lRespublica:
@hartwork Well, I have discussed this with other members of the community. We agreed that it would still be better to build from .tar.gz. Probably PR should have been closed.

Thank you for your cooperation!

hartwork:
@CausaPrincipalis71 so you'll do the `and not _path_is_child_of('/usr/src', __file__)` patching in ALT then, is that the plan?

PR #2050, opened 2023-04-19
build-deb.yml: Avoid zstd compression for compatibility (fixes #2049)

### LINKED ISSUE #2049: PyChess Debian packages should avoid zstd compression for compatibility
ubuntu-latest is now ubuntu 22.04

ubuntu has an extension to apt that uses tar.zst to compress file in debs. this is used by default by dpkg when packaging debs in 22.04

this prevents all debs from installing on debian/raspbian

you need to switch to building with ubuntu-20.04 or ubuntu-18.04 which do not use this functionality when packaging debs

https://github.com/Botspot/pi-apps/actions/runs/4738073287/jobs/8411536446#step:4:2155

alternativly you can specify the compression method as gzip through a dpkg command like option, debhelper override, or whatever method you use to build these debs

hartwork:
Fixes #2049

CC @theofficialgman

### DIFF
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -27,6 +27,7 @@ jobs:
         sudo apt-get install --no-install-recommends --yes -V \
             debhelper \
             dh-python \
+            file \
             gettext \
             gir1.2-gdkpixbuf-2.0 \
             gir1.2-glib-2.0 \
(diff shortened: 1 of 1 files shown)


gbtami (approved):

gbtami:
Thx!

theofficialgman:
this is pretty nasty so I looked into why you do this
turns out the python debian helper you use just really isn't great and doesn't allow you to pass in any arguments to dpkg-buildpackage that it hardcodes in its code
https://github.com/astraw/stdeb/blob/bc37aa183238044267478e610dee755db5b96280/stdeb/command/bdist_deb.py#L59
that is unfortunate because it is as simple as passing in `--compression=xz` to change the compression. that functionality should probably be requested as a new feature from upstream stdeb

hartwork:
@theofficialgman `DPKG_DEB_COMPRESSOR_TYPE` didn't work, so re-compressing was the obvious candidate, I don't consider it "nasty", definitely better than downgrading CI to an older version of Ubuntu. Personally I believe in compiling from source over use of binary packages wherever possible, so I won't join a fight with stdeb.

PR #2084, opened 2023-07-02
Update pychess.po part 1

M-DinhHoangViet:
Add trangle Vietnamesse part 1


gbtami:
Thx for your work, but don't edit .po files in git. Use transifex please. We pull translations from there.
See https://github.com/pychess/pychess/wiki/TransifexTranslates

PR #2119, opened 2023-10-06
Add support for Python 3.12

hartwork:


gbtami:
Seems this needs https://github.com/marcelotduarte/cx_Freeze/issues/2153

hartwork:
> Seems this needs [marcelotduarte/cx_Freeze#2153](https://github.com/marcelotduarte/cx_Freeze/issues/2153)

I'm not sure I want to vote for `--extra-index-url https://marcelotduarte.github.io/packages/` to use their own versions of third party wheels. It will have fewer eyes using and auditing than ideal.

gbtami:
Yea, we better to wait for official pypi releases.

gbtami:
https://pypi.org/project/cx-Freeze/ 7.0.0 is out now.

hartwork:
@gbtami finally green CI, please merge **without** squashing if happy with the changes, semantic commit cuts really matter here. Thanks!

hartwork:
@gbtami thanks for the merge! Should we release 1.0.5 off `master` with Python 3.12 support?

gbtami:
Yes, absolutely!

hartwork:
> Yes, absolutely!

@gbtami I have create pull request #2197 for that and rebased it on top of all other merged branches now. The tests seem to be a bit flaky today — not sure why — but after a few re-runs the CI seems to go back to green :thinking: 

PS: Thanks for the invite to the GitHub org. I would turn my membership there to public but it seems like everyone else there is private, and me then being the only public member would look weird, so I have put myself back to private just because of that. Any ideas?

gbtami:
Haha, I never noticed this private/public thing is exists at all. Possible this is the case for the other members as well idk :)
Anyhow, I changed mine to public in pychess and in Fairy-Stockfish as well now.
Thx for noticing it.

hartwork:
@gbtami okay, I changed mine back to public now too. For 1.0.5, what we need more: a simple Git tag (if we want to match 1.0.4), a GitHub release page, with binaries from the CI added. I think you did all of that for 1.0.4 but I could mis-remember. How should we continue?

gbtami:
I downloaded the generated .msi to test it on my Win10, but unfortunately it failed to start :(
msi

Interestingly when I run
`python setup.py bdist_msi`
in my local MSYS2 install (it has older package versions of everything compared to current) it works OK.
I can imagine this is related to the new cx_freeze...

gbtami:
Also after my latest commit some github action were not successful :(

hartwork:
> I downloaded the generated .msi to test it on my Win10, but unfortunately it failed to start :(

@gbtami I'm I cannot offer help with that beyond Googling for it.



> Also after my latest commit some github action were not successful :(

@gbtami the Ubuntu mirror of Microsoft had issues of temporary nature apparent. I re-ran CI just now and your latest commit 67d048f42a4044e3fa484cd8bce829b0e3622b9e is all green now.

masoudd:
I tested the msi installer from build artifacts of [this ci run](https://github.com/pychess/pychess/actions/runs/8820031832) from commit e8e9c548da79cf82fd84492e6f9d09664acd6634
And it installs successfully but running it gives this error and terminates:
Screenshot 2024-04-24 225352
Using up to date windows 10

hartwork:
@masoudd @gbtami after a look at https://github.com/python/cpython/blame/main/Lib/asyncio/windows_events.py my impression is that the CPython 3.11 in use there is incomplete (and hence broken). Or it fails to find the file in the right place? Is that something that cx_Freeze would influence?

masoudd:
I think that is something wrong with cx_Freeze, I did some digging and came up with this:
add `"_overlapped"` to array in [line 317 in setup.py](https://github.com/pychess/pychess/blob/master/setup.py#L317)

So line 317: `"includes": ["gi", "_overlapped"],`

I built an msi file on my local msys install and installed it. Now running pychess gives this error:
new_error

Which suggests something is wrong with either pygobject or cx_Freeze. I can run pychess from a copy of the master branch checked out in my msys install, so I think that would rule out pygobject.

hartwork:
@masoudd interesting, thanks for sharing! I hope that @gbtami can take that hint and figure something out, I don't have access to Windows myself outside of CI and MinGW, and have not have for 10+ years.

gbtami:
I opened https://github.com/marcelotduarte/cx_Freeze/issues/2356

hartwork:
@gbtami that's great, thank you! And we have a fix from upstream and in MSYS already. Does the build at https://github.com/pychess/pychess/actions/runs/8837443869 happen to already have that fix and work?

gbtami:
Seems the fixed package is not available yet.
https://packages.msys2.org/package/mingw-w64-x86_64-python-cx-freeze?repo=mingw64

hartwork:
@gbtami I see! Let's check again in a few hours maybe.

gbtami:
I got this link to check :) https://packages.msys2.org/queue

hartwork:
@gbtami nice! It's past the queue now if I read the page right. I triggered a rebuild to latest `master` at https://github.com/pychess/pychess/actions/runs/8837952115 , let's see.

hartwork:
@gbtami you won't like it, I'm afraid:

```
running build_exe
  File "D:/a/pychess/pychess/setup.py", line 365, in <module>
    setup(
  [..]
    module.hook(self)
  File "D:/a/_temp/msys64/mingw64/lib/python3.11/site-packages/cx_Freeze/hooks/__init__.py", line 605, in load_subprocess
    module.exclude_names.update(exclude_names)
                                ^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'exclude_names' where it is not associated with a value
```

gbtami:
Opened https://github.com/marcelotduarte/cx_Freeze/issues/2362

masoudd:
I don't fully understand cx_Freeze but this pull request seems to be the fix: https://github.com/marcelotduarte/cx_Freeze/pull/2363

Edit: I used the change in the above pull request and it builds the msi file. But installing it and running pychess gives the error I posted above: `assert g_type != TYPE_NONE`. 
new_error
They know about it: https://github.com/marcelotduarte/cx_Freeze/issues/2361

hartwork:
@gbtami @masoudd the Windows CI is green again just now, could you try the artifact from the re-build at https://github.com/pychess/pychess/actions/runs/8855111639 ?

gbtami:
It produces the same AssertionError masoudd pasted above :(

hartwork:
@gbtami okay :disappointed:

hartwork:
@gbtami should we tag and release 1.0.5 to https://github.com/pychess/pychess/releases without Windows binaries OR wait for working Windows binaries and then tag and release 1.0.5?

gbtami:
Yes, we can tag and release without github CI .msi I think. (Edit: done)
Possible I will upload .msi from my older (but working) local MSYS2 install.

hartwork:
@gbtami thanks for the tag and release! One key file seems missing still: `pychess-1.0.5.tar.gz` with `eco.db` pre-generated as with 1.0.4: https://github.com/pychess/pychess/releases/tag/1.0.4 .

hartwork:
@gbtami PS: I noticed when trying to bump the package in Gentoo.

hartwork:
@gbtami PPS: it's generated as `pychess_1.0.5.orig.tar.gz` as part of the Debian CI, if I remember correctly: https://github.com/pychess/pychess/actions/runs/8901837037 . The tarball is contained in that artifact zip.

gbtami:
Thx. Added.

hartwork:
> Thx. Added.

@gbtami thank you! I dared [renaming](https://github.com/pychess/pychess/releases/tag/1.0.5) `pychess_1.0.5.tar.gz` (with underscore) to `pychess-1.0.5.tar.gz` (with hyphen) for consistency with 1.0.4, I hope that's in your interest also and that most packagers try packaging after the rename rather than in the time window before.

hartwork:
@gbtami PS: https://github.com/gentoo/gentoo/commit/2068e501018a8bff395c064e8b85c4af4a7048be just put PyChess 1.0.5 into Gentoo

PR #2148, opened 2023-12-27
Disable engine auto detection for 'spike'

masoudd:
**spike** is an [unrelated software in Archlinux](https://github.com/riscv-software-src/riscv-isa-sim) repository. I searched for **spike** in Debian, Ubuntu and Fedora package indexes, and they also don't package the [spike chess engine](http://spike.lazypics.de/index_en.html).
So I think if the user is going to install the spike engine manually they would also add it to pychess themselves.

See [this comment](https://gitlab.archlinux.org/archlinux/packaging/packages/pychess/-/blob/1.0.4-3/PKGBUILD?ref_type=tags#L16) from the Archlinux pychess packager. pychess mistakes `/usr/bin/spike` for a chess engine.


gbtami:
Thx!

PR #2187, opened 2024-04-21
Added intuitive sort by elo for databse game list treeview

### LINKED ISSUE #2186: Sorting the game list by elo is in lexicographic order
Hello,

When browsing my games in PyChess, and asking to sort them by elo, I noticed that ratings within the 3 digits range come before the ratings within the 4 digits... I then understood where the issue comes from, the `TreeView` widget is given the elo data as a string, hence the lexicographic order.

before_sort_fix

The first solution I tried was changing a few lines to the widget so that it considers the elo as an `int`, but then it requires a default value for the games with undetermined elo (like -1`...). I managed to get around the issue like that, but that might not be the optimal solution.

So I found in the GTK doc a way to implement a custom sort function, but I couldn't manage to make it work in the app.

I added that function in `lib/pychess/perspectives/database/gamelist.py` :
  
 ```python
def compare_elo(self, treesortable, row1, row2, user_data):
    sort_column, _ = treesortable.get_sort_column_id()

    elo1 = treesortable.get_value(row1, sort_column)
    elo2 = treesortable.get_value(row2, sort_column)

    isdig1 = elo1.isdigit()
    isdig2 = elo2.isdigit()

    if isdig1 and isdig2:
        dig1 = int(elo1)
        dig2 = int(elo2)
        return (dig1 > dig2) - (dig1 < dig2)
    elif isdig1:
        return 1
    elif isdig2:
        return -1
    else:
        return (elo1 > elo2) - (elo1 < elo2)
```

but the line

```python
self.liststore.set_sort_func(2, self.compare_elo, None) # or self.modelsort
```

doesn't seem to do anything when putting it inside the `__init__`, even though the function is called and returns the expected value.

I would love to make a PR once I solve this issue :)

Thanks!

tedrhuyagonall:
This adds a custom sort function to make sorting by elo in the treeview more intuitive: `950` will no longer be greater than `2400`, for example. If the elo is just `?`, or empty, it will still be sorted in lexicographic order.

This should fix https://github.com/pychess/pychess/issues/2186.

### DIFF
--- a/lib/pychess/perspectives/database/gamelist.py
+++ b/lib/pychess/perspectives/database/gamelist.py
@@ -60,6 +60,8 @@ class GameList(Gtk.TreeView):
         self.modelsort = Gtk.TreeModelSort(self.liststore)
 
         self.modelsort.set_sort_column_id(0, Gtk.SortType.ASCENDING)
+        self.modelsort.set_sort_func(2, self.compare_elo, 2)
+        self.modelsort.set_sort_func(4, self.compare_elo, 4)
         self.modelsort.connect("sort-column-changed", self.sort_column_changed)
         self.set_model(self.modelsort)
         self.get_selection().set_mode(Gtk.SelectionMode.BROWSE)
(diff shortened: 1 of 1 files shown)


gbtami:
Thx!

Issue #2233, opened 2024-06-23
telnetlib is removed in Python 3.13


jelly:
See [whats new](https://docs.python.org/3.13/whatsnew/3.13.html#removed-modules-and-apis) and the usage [here](https://github.com/pychess/pychess/blob/63325b4abe121469f7daf1e68f9734921208f4bd/lib/pychess/ic/TimeSeal.py#L3).

AdamWill:
seems like it only uses three constants from telnetlib? maybe just vendor those?

gbtami:
Fixed via https://github.com/pychess/pychess/pull/2235

PR #2235, opened 2024-06-25
TimeSeal.py: make IAC_WONT_ECHO a literal as telnetlib is gone (#2233)

AdamWill:
telnetlib was removed in Python 3.13. As we only used it to create a bytestring from these constants, let's just make it a literal instead.


gbtami:
Thx!

PR #2359, opened 2025-07-21
Fix GHA workflow so test failure is properly caught

AdamWill:
The current GHA "Run the test suite" workflow will virtually never fail, because run_tests.py never exits anything but 0. It effectively runs the tests, but ignores the results. The tests have actually been failing for months - likely since 983e138 - but nobody noticed because of this problem.

This changes run_tests.py so it exits 1 if the tests fail, and tweaks run-tests.yml so we will still kill xvfb before exiting if the tests fail.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -77,6 +77,8 @@ jobs:
         xvfb_pid=$!
 
         cd testing
-        ./run3 run_tests.py
+        ./run3 run_tests.py || touch FAILED
 
         kill -s SIGTERM "${xvfb_pid}"
+        [ -f FAILED ] && exit 1
+        exit 0
--- a/testing/run_tests.py
+++ b/testing/run_tests.py
@@ -1,3 +1,4 @@
+import sys
 import unittest
 
 modules_to_test = (
@@ -45,4 +46,7 @@ def suite():
 
 
 if __name__ == "__main__":
-    unittest.TextTestRunner(verbosity=2).run(suite())
+    ret = unittest.TextTestRunner(verbosity=2).run(suite())
+    if ret.wasSuccessful():
+        sys.exit(0)
+    sys.exit(1)


AdamWill:
The test failure is expected - that's what this PR fixes. If you check the actual logs of test executions on other recent PRs, the tests failed the same way in all of them, but the result was considered a success due to this bug.

.github/workflows/run-tests.yml
```diff
@@ -77,6 +77,8 @@ jobs:
         xvfb_pid=$!
 
         cd testing
-        ./run3 run_tests.py
+        ./run3 run_tests.py || touch FAILED
 
         kill -s SIGTERM "${xvfb_pid}"
+        [ -f FAILED ] && exit 1
+        exit 0
```
QuLogic, line 84: I believe this is bash, so can probably use `trap` (and also change `set -x` to `set -ex` to error directly)?
```suggestion
        xvfb_pid=$!
        trap "kill -s SIGTERM ${xvfb_pid}" EXIT

        cd testing
        ./run3 run_tests.py
```

AdamWill: I mean, sure, is it *better* though? I kinda like simple. :P

gbtami:
Thx.

hartwork:
@gbtami while I support the general direction, please never merge things that make the CI red. When things make the CI red, they should bring the needed fixed to make it green again piggy backed: red CI is a disaster because now all new pull requests will be red until the issue is fixed. I'm not surprised if a revert is not welcome, but in some sense red CI is even worse than overly optimistic green CI, because (1) it makes merging with red the new norm and (2) makes introduction of new errors harder to see. I really hope the plan is to [get CI fully fixed in the next few days then](https://blog.hartwork.org/posts/my-approach-to-code-review/#B3) then. Just my 2 cents. Thanks for the work on this issue to @AdamWill.

gbtami:
I absolutely agreed. I was thinking the latest two PRs together will make CI green.
Possible I was misunderstanding it. My bad.
I don't know how to proceed now, because if we revert them, we just get the "fake green" CI again.

hartwork:
@gbtami cool! Let me run a quick experiment in CI before suggesting something concrete.

CC @AdamWill

AdamWill:
Yeah, no, sorry, I should have made that more clear: the CI would stay red until https://github.com/pychess/pychess/issues/2360 is fixed. I do not have a fix for that. https://github.com/pychess/pychess/pull/2361 fixes the tests on Python 3.14 *other than* the problem in #2360 - the tests pass on 3.14 if you apply that change on top of the commit before the one that caused 2360.

I'm hoping you or @hartwork can figure out #2360, since you're more familiar with how the asyncio stuff is *intended* / *expected* to work.

hartwork:
I'm not familiar with the very asyncio bits in pychess (and am still learning about asyncio) *but* there is a candidate pull request with a fix at #2362 now. I'm looking forward to review :pray:

Issue #2360, opened 2025-07-21
analysis and learn tests fail due to lack of a running asyncio loop


AdamWill:
Tests in analysis.py and learn.py fail due to no running asyncio loop:
```
======================================================================
ERROR: test1 (analysis.CECPTests.test1)
Test analyzing in forced mate situations
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/pychess/pychess/testing/analysis.py", line 83, in setUp
    self.engineA, self.analyzerA = self._setupengine(ANALYZING)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/pychess/pychess/testing/analysis.py", line 60, in _setupengine
    analyzer = CECPEngine(engine, WHITE, 2, 0)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/pychess/pychess/lib/pychess/Players/CECPEngine.py", line 174, in __init__
    self.parse_line_task = asyncio.create_task(self.parseLine(self.engine))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/asyncio/tasks.py", line 417, in create_task
    loop = events.get_running_loop()
           ^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: no running event loop

======================================================================
ERROR: test2 (analysis.CECPTests.test2)
Test analyzing in promotion situations
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/pychess/pychess/testing/analysis.py", line 83, in setUp
    self.engineA, self.analyzerA = self._setupengine(ANALYZING)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/pychess/pychess/testing/analysis.py", line 60, in _setupengine
    analyzer = CECPEngine(engine, WHITE, 2, 0)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/pychess/pychess/lib/pychess/Players/CECPEngine.py", line 174, in __init__
    self.parse_line_task = asyncio.create_task(self.parseLine(self.engine))
[...]

PR #2361, opened 2025-07-21
Adjust test usage of get_event_loop() for Python 3.14 changes

AdamWill:
As of Python 3.14, get_event_loop() no longer automatically sets up a new loop if there currently isn't one, instead it raises a RuntimeError. This adjusts several usages of get_event_loop() in the tests to work with the change. It also adds a loop setup step to learn.py - currently it doesn't have one, but calls asyncio.create_task, so it is implicitly relying on an earlier test having set up an event loop. If you run learn.py alone, it fails because of this.

### DIFF
--- a/testing/analysis.py
+++ b/testing/analysis.py
@@ -78,7 +78,11 @@ class CECPTests(EmittingTestCase):
         self.assertEqual(results, ([(ply, moves, score, depth, nps)],))
 
     def setUp(self):
-        self.loop = asyncio.get_event_loop()
+        try:
+            self.loop = asyncio.get_event_loop()
+        except RuntimeError:
+            self.loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(self.loop)
 
         self.engineA, self.analyzerA = self._setupengine(ANALYZING)
         self.engineI, self.analyzerI = self._setupengine(INVERSE_ANALYZING)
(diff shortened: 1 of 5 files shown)


AdamWill:
Note: tests on this will fail because of https://github.com/pychess/pychess/issues/2360 (but this won't be shown properly until https://github.com/pychess/pychess/pull/2359 is merged), but I tested these changes on 1.0.3 to avoid that problem, and they allow the tests to pass with Python 3.14. If you run 1.0.3 tests on Python 3.14 without these changes, they fail.

gbtami:
Thx.

hartwork:
`master` is [red after merge](https://github.com/pychess/pychess/actions/runs/16442279673/job/46465691448) with related errors so something about this change is not complete (or I am missing something)?

CC @gbtami @AdamWill

AdamWill:
sorry, I obviously didn't explain clearly enough: this doesn't fix #2360, so yeah, it was expected that CI would still be red after this.

This fixes a different problem: if you go to 1.0.3 or the commit right before https://github.com/pychess/pychess/pull/1995 (the one that caused #2360), the tests pass if run on Python up to 3.13, but fail on Python 3.14. This makes the tests pass on Python 3.14 *if #2360 is taken out of the equation*. I couldn't figure out a good fix for #2360 yesterday - I was hoping one of you two could.

PR #2364, opened 2025-07-23
Ics stream reader protocol

### LINKED ISSUE #2299: broken after upgrade to Fedora 41 / python 3.13
Hello,
  As is now custom, pychess has become unusable after updating my system. 
  Fedora common install uses 1.0.3. Trying to start a FICS game will abort and disconnect upon the first move. 
  "AttributeError: 'ICSStreamReaderProtocol' object has no attribute '_stream_writer"
is the terminal error. 
 Is there a current fc41 1.0.5 rpm, some way to use this with python 3.13 or any sort of workaround?
  Thanks as always for keeping up with this. It remains for preferred FICS client.
  -A

gbtami:
Fixes https://github.com/pychess/pychess/issues/2299

### DIFF
--- a/lib/pychess/ic/TimeSeal.py
+++ b/lib/pychess/ic/TimeSeal.py
@@ -90,7 +90,7 @@ class ICSStreamReaderProtocol(asyncio.StreamReaderProtocol):
 
             if self.timeseal:
                 for i in range(g_count):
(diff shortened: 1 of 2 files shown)


AdamWill:
gah, well that was easy! why didn't I notice that was available. :D testing

PR #2366, opened 2025-07-24
Fix RuntimeWarning: coroutine 'Event.wait' was never awaited

gbtami:


.github/workflows/pre-commit.yml
```diff
@@ -21,5 +21,5 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
     - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
       with:
-        python-version: 3.13
+        python-version: 3.12
```
hartwork: @gbtami this :arrow_up: seems mistaken?

gbtami (commented):
@hartwork I use Ubunu 24.04 LTS which ship Python 3.12, if we change this workflow to use 3.13 I can't use pre-commit locally.
Or maybe I just missing something here?

hartwork:
@gbtami the Python version there is only affecting CI not local development and nothing more than what the pre-commit process itself is run by. It's not the target language, which would still be Python 3.9.

gbtami:
OK, at least the pre-commit is green now :)

hartwork:
> OK, at least the pre-commit is green now :)

@gbtami cool! :+1: I can add [CI that keeps pre-commit hooks up to date for us via pull requests](https://github.com/hartwork/git-delete-merged-branches/blob/HEAD/.github/workflows/pre-commit-detect-outdated.yml) after this is merged, if you like.

gbtami:
Sure, absolutely!

hartwork:
@gbtami here you go: #2367

PR #2384, opened 2025-09-13
Add blindfold modes, hidden white and hidden black

Endothelin:
I added two additional modes that might be a helpful step towards playing in full blindfold mode. In "hidden white" all white figurines are hidden and in "hidden black" all black figurines are hidden.


gbtami:
Thx a lot! Telling the truth all of these blindfold "variants" should be implemented in settings because they are just visual things only, kind of similar to piece sets and not real chess variants. But this is how it is now. It was my bad choice long years ago.
Anyhow if you wanna implement them in the settings instead feel free to create a new PR for that.

Endothelin:
I have to admit that I implemented the modes in that simplest possible way, but agree with your suggestion to move the blindfold modes into the settings. Please see #2385 for my proposal.

PR #2385, opened 2025-09-14
Move blindfold variants into settings

Endothelin:
As mentioned in your comment in #2384, this moves the blindfold modes into the settings. I tried to remove the old code as well as I could, but there is probably some code left that could be removed (for example, the translations).


gbtami:
Thx a lot! Now you should find the way to immediately redraw the pieces when any of the new settings changes similar to how the other settings work. Now I have to close the dialog to see the effect of the changes.
Somewhere in https://github.com/pychess/pychess/blob/master/lib/pychess/widgets/preferencesDialog.py IIRC

Endothelin:
Sorry for missing that. With the new changes the display is updated instantly from the settings.

PR #2404, opened 2025-10-12
Move config into config dir

Endothelin:
The configuration was stored in the cache directory instead of the config directory. To me this seems to be the wrong directory and clearing the cache loses the configuration.

### DIFF
--- a/lib/pychess/Database/model.py
+++ b/lib/pychess/Database/model.py
@@ -22,7 +22,7 @@ from sqlalchemy.exc import OperationalError
 # from sqlalchemy.sql.expression import Executable, ClauseElement, _literal_as_text
 
 from pychess.System.Log import log
-from pychess.System.prefix import addUserCachePrefix
+from pychess.System.prefix import addUserConfigPrefix
 
 
 @event.listens_for(Engine, "before_cursor_execute")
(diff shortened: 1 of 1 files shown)


hartwork:
@gbtami any thoughts?

gbtami:
I'm not sure about this. 20180221.sqlite is not really a configuration file. It is an empty sqlite database file used as skeleton when the user creates a new chess database. If you clear the cache it will be re created next time when you start pychess again.

Endothelin:
You are right it, it is recreated when deleted. However, the user looses the settings that are stored in it.

Edit: I wanted to add an example of a setting that is lost when the cache folder is cleared, but no settings where lost, when I deleted the file. But I loose (some) pychess settings every time I restart my PC. I must have somehow confused something when creating this change. I will try to figure out why I loose settings, but at the moment I do not know how to figure out how to do that.

gbtami:
What is the exact setting you lose when you restart your PC?

Endothelin:
I just tested it on two different PCs and did not loose any settings. However, I am pretty sure I lost settings sometime in the past as I remember setting them repeatedly. As the problem no longer exists. I withdraw this Pull request. I am sorry for wasting your time.

Endothelin:
When starting pychess today, my settings were reset again. I am not sure what caused it, but I wanted to document it anyway. The "Tip of the day" and "Ask for permissions" popups appeared again. Sounds and auto save were activated again. However, my layout of the views is still as I configured it.

PR #2414, opened 2025-10-27
Fix CI / Workaround issues with pip >=25.3

hartwork:
The import `from pychess.Savers.pgn import PGNFile` in file `setup.py` internally does an `import pexpect` which started failing with error `ModuleNotFoundError: No module named 'pexpect'` because pip >=25.3 no longer respects the outer virtualenv during `pip3 install .`. This is a workaround only. Likely a migration to `pyproject.toml` is needed for a proper fix.


gbtami (approved):
Thx!

Issue #2242, opened 2024-07-20
Stockfish processes don't close after exiting PyChess


vantiki:
I recently came across an interesting point when using the Stockfish engine together with PyChess.

The engine was connected using the standard method: Edit -> Engines and a directory with Stockfish version 16.1 was added. 

At the start of each new game at least 1 engine process is started. However, PyChess does not kill old processes and the number of them increases with each new game.

Of course, you can manually search for processes via ps aux | grep stockfish and then kill them with kill -9, but it is not pleasant to do such manipulations after each finished game.

That's why I ask you to fix this bug. Let PyChess at the end of the game kill the process or, at least, not open new ones, so as not to overflow the computer's RAM for nothing.

Operating system: Linux (Gentoo)
PyChess version: 1.0.5
Stockfish version: 16.1

masoudd:
What is your python version?
This bug keeps resurfacing since a few years ago. It's a mix of multiprocess and asyncio logic and I fear I'm not smart enough to find it. See also #2210 #1949 #1918 #1807

gbtami:
Fixed via https://github.com/pychess/pychess/commit/78d20f1beb0b4bd2943cf67ea4e447de32fe4b13

app-exit/analyzer part is fixed but ended game tabs still retain player engines until close
    because of undo semantics

Issue #2210, opened 2024-05-11
Analysis engine keeps running in the background even after the game ends


60k41p:
PyChess version 1.0.5 (official .deb downloaded from releases/), running on Debian 12 x86_64
Stockfish 16.1 configured as analysis engine.

Issue observed playing both as white and black side and against different engines as opponents (pychess engine & monchester).

Steps to reproduce:
1. Start a new game against an engine, time control: unlimited, variant: normal chess
2. Play the match normally. Analysis engine confirmed to be running in the background
3. Win, lose or draw match - the analysis engine is still running in the background
4. Start a new match via Game -> New Game
5. A new instance of the analysis engine is process is launched, the previous game's instance is still running.

Closing the match tab for the previous game either before step 4 or after does not make any difference, the engine is still running in the background.

Only when PyChess application is closed does it terminate the analysis engine process(es).

Preferences set as follows:
![image](https://github.com/pychess/pychess/assets/9840837/ea87f898-34ed-4690-aa66-b2d57097d367)

60k41p:
Similar to https://github.com/pychess/pychess/issues/1807 & https://github.com/pychess/pychess/issues/1918.

60k41p:
problem doesn't occur with [berserk](https://github.com/jhonnold/berserk)

gbtami:
Fixed via https://github.com/pychess/pychess/commit/78d20f1beb0b4bd2943cf67ea4e447de32fe4b13

PR #2468, opened 2026-04-09
Resume stopped engine process before terminate

scouarn:
Engines wouldn't terminate correctly after closing a game and stay until the application is closed. If the process was stopped with SIGSTOP, SIGTERM is not handled until a SIGCONT is received. Adding a `self.resume()` call seams to fix it on my archlinux with stockfish and PyChess.py.

This should fix the following issues: #2242 #2210

Just to let you know, I'm not 100% sure it is enough since I'm not very good at Python subprocesses and async io. I was initially testing with my custom engine that creates some processes of its own. The fix works When removing the subprocesses creation, but with them on it still hangs half the time and SIGTERM is only received when the app is closed. Adding `self.proc.stdin.close()` along with the `self.resume()` fixes it. I didn't included it in the commit since it may be some bad signal handling on my part.

### DIFF
--- a/lib/pychess/System/SubProcess.py
+++ b/lib/pychess/System/SubProcess.py
@@ -155,6 +155,7 @@ class SubProcess(GObject.GObject):
         if self.write_task is not None:
             self.write_task.cancel()
         self.read_stdout_task.cancel()
+        self.resume()
         try:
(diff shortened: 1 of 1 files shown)


gbtami:
Thx a lot!

PR #2469, opened 2026-04-11
pre-commit: Make Github Dependabot keep pre-commit hooks up to date

hartwork:
@gbtami this will resolve the "close, reopen, mark as ready to review" loop and also protect us from moving Git tags


gbtami (approved):

gbtami:
Looks good.

PR #2472, opened 2026-04-14
`setup.py`: Drop a Python 3.9 leftover (follow-up to #2406)

hartwork:
Follow-up to commit 3a4bfbaf4250b4470ded727813d9f17c9b1f5422 from pull request #2406


hartwork:
@gbtami PS: Motivation was to hopefully unblock the update of pytest to 9.0.3 that seems to be blocked by this, judging from https://github.com/pychess/pychess/actions/runs/24378418613/job/71196572908 . Let's see…

PR #2483, opened 2026-04-24
Modernize project tooling

gbtami:


pyproject.toml
```diff
@@ -0,0 +1,111 @@
...
+builtins = ["_", "ngettext", "reload"]
+extend-exclude = [
+    "build",
+    "dist",
+    "doc/createDoc",
+    "lib/pychess.egg-info",
+]
+
+[tool.ruff.lint]
+select = ["E", "F"]
```
hartwork, line 79: @gbtami could this be missing ["UP"](https://docs.astral.sh/ruff/rules/#pyupgrade-up)?

gbtami:
Good idea. 
I tried: `uv run ruff check . --select UP --fix`
It was able to fix lots of old style string formatting lines, but left huge amount of them as not fixable automatically. Possible those needs fixing by hands which seems a bit too much for me.
What do you think?

hartwork:
@gbtami I see! I believe the options are:
- a) Mass-fix manually once
- b) Mass-mark as "noqa" every single unfixed case
- c) Enable some of `UP` but not all, i.e. add all failure classes to `ignore` *after* auto-fixing
- d) Add all files with violations to an exclude list.

(c) is the only way forward that is both cheap and without considerable downsides. How about that?

gbtami:
Sounds good.

hartwork:
@gbtami should I do a pull request or are you already at one?

gbtami:
Please do.

PR #2495, opened 2026-05-20
perf: ~21% engine speedup via quiescent eval cache and bit_count() popcount

samqbush:
## Summary

Profile-directed performance optimizations to the chess engine search. Two independent improvements were applied, yielding a combined **~21% speedup** (8.57s → 6.78s on a standard benchmark set, depth 1–4).

### 1. Quiescent search evaluation cache (`lsearch.py`)

The quiescent search calls `evaluateComplete()` at every leaf node for stand-pat evaluation. Profiling revealed that **53.7% of those calls evaluate positions already seen** earlier in the same search tree (transpositions via different capture sequences).

A lightweight Zobrist-keyed dict cache (`_eval_cache`) was added inside `quiescent()`:
- On a cache hit, the stored score is returned directly — no re-evaluation.
- On a miss, `evaluateComplete()` is called and the result is stored.
- The cache is cleared at the start of each root `alphaBeta()` call (`ply == 0`) to prevent unbounded growth across iterative deepening steps.

This is correct because `evaluateComplete(board, color)` is a pure function of board state, and `board.hash` (Zobrist) already encodes both position and side-to-move.

**Impact:** Cuts ~128K unnecessary `evaluateComplete` calls per search, saving the majority of the 6.5s cumulative time that function consumed.

### 2. Replace `bin(x).count("1")` with `x.bit_count()` (multiple files)

25 occurrences of `bin(x).count("1")` were replaced with `x.bit_count()` across `leval.py`, `strateval.py`, `lsearch.py`, `lmovegen.py`, `losers.py`, `suicide.py`, and `EndgameTable.py`.

`bin(x).count("1")` allocates a Python string and iterates over its characters. `int.bit_count()` (Python 3.10+) calls the native popcount instruction directly, running ~4× faster per call.

## Profiling methodology

```
cProfile, sorted by tottime, 2 positions at depth 1–4
```

Key findings before optimization:
| Function | Calls | Cumulative time |
|---|---|---|
| `evaluateComplete` | 278,311 | 6.5s |
| `applyMove` | 317,329 | 2.2s |
| `staticExchangeEvaluate` | 377,639 | 2.4s |
| `iterBits` | 6,801,013 | 0.8s |

## Benchmark results

| | Time | Nodes/sec |
|---|---|---|
| Before | 8.57s | 37,000 n/s |
| After | 6.78s | 47,000 n/s |
| **Speedup** | **~21%** | **~27% more nodes/sec** |

Search results (scores and best moves) are identical before and after — no behavioral change.

## Tests

All existing tests pass:

```
Ran 17 tests in 0.28s
OK
```

Tests covered: `alphabeta`, `eval`, `bitboard`, `Board`, `atomic`, `suicide`, `losers`.

### DIFF
--- a/lib/pychess/Utils/EndgameTable.py
+++ b/lib/pychess/Utils/EndgameTable.py
@@ -23,7 +23,7 @@ class EndgameTable(GObject.GObject):
         self.providers = providers
 
     def _pieceCounts(self, board):
-        return sorted([bin(board.friends[i]).count("1") for i in range(2)])
+        return sorted([(board.friends[i]).bit_count() for i in range(2)])
 
     def scoreGame(self, lBoard, omitDepth=False, probeSoft=False):
         """Return result and depth to mate. (Intended for engine use.)
--- a/lib/pychess/Utils/lutils/leval.py
+++ b/lib/pychess/Utils/lutils/leval.py
@@ -337,24 +337,24 @@ def cacheablePawnInfo(board, phase):
                 not (passedPawnMask[opcolor][i] & ~fileBits[cord & 7] & pawns)
                 and board.arBoard[i] != PAWN
             ):
-                n1 = bin(pawns & moveArray[opptype][i]).count("1")
-                n2 = bin(oppawns & moveArray[ptype][i]).count("1")
+                n1 = (pawns & moveArray[opptype][i]).bit_count()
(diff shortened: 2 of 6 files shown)


gbtami:
Looks like a very nice enhancement. Thx a lot!

PR #2514, opened 2026-06-25
Add language selection to Preferences (#2500)

### LINKED ISSUE #2500: Add a Language Selection in Preferences Menu (feature request)
Seems like there's currently no way to switch languages from within the app. This would be a useful option in the preferences menu for people who are bilingual or who wish to view other translations without changing their system settings.

philipmk42:
Closes #2500.

Adds a Language dropdown to the General tab of Preferences so users can choose a UI language without changing their system locale.

Changes:
- glade/PyChess.glade: add a Language label and `langcombo` to the General Options box.
- conf.py: add a `General/langcombo` default, where an empty string means follow the system locale.
- preferencesDialog.py: discover the catalogs already shipped under `lang/`, populate the combo with readable language names, and persist the chosen locale code (not a positional index, so it stays stable if the language list changes).
- pychess launcher: read the setting before `gettext.install()` and set `LANGUAGE` so the chosen catalog loads.

Behavior:
As discussed in the issue, this applies on the next restart rather than switching live, since translations are bound when widgets are built and the windows are cached singletons. The combo tooltip notes the restart requirement. A true live switch would be a larger, separate effort.

Testing:
ruff lint and format pass. Loading the edited glade in a real GTK3 GtkBuilder succeeds, with the language widgets created, correctly typed, and placed in the General Options box. The catalog discovery, combo population, and config persistence (selecting a language writes its locale code, "System default" writes an empty string) were verified against the actual code under Xvfb. I was not able to do a full interactive run with compiled catalogs, so a visual confirmation of the restart switch on Linux would be welcome.

### DIFF
--- a/glade/PyChess.glade
+++ b/glade/PyChess.glade
@@ -1420,6 +1420,45 @@ Ticked : the linear scale focuses on a limited range around the null average. It
                                         <property name="position">12</property>
                                       </packing>
                                     </child>
+                                    <child>
+                                      <object class="GtkHBox" id="langbox">
+                                        <property name="visible">True</property>
+                                        <property name="can-focus">False</property>
+                                        <property name="spacing">6</property>
+                                        <child>
+                                          <object class="GtkLabel" id="langlabel">
+                                            <property name="visible">True</property>
+                                            <property name="can-focus">False</property>
+                                            <property name="halign">start</property>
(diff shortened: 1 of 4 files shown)


philipmk42:
Thanks for the review and merge! Happy to open a follow-up issue scoping a true live-switch refactor if there's interest down the line.

PR #2533, opened 2026-08-18
Drop broken ChessPastebin.com and Gameknot.com support

hartwork:
In reaction to https://github.com/pychess/pychess/actions/runs/32028858953/job/95395448236?pr=2532#step:6:4056

CC #2532


gbtami:
Thx.

PR #2536, opened 2026-08-20
Limit CI runtime and package installation runtime

hartwork:
Sibling of https://github.com/libexpat/libexpat/pull/1311

@gbtami I believe we have seen CI runs here also where CI timed out after many hours of trying `sudo apt-get install` with a turtle-speed slow mirror. This pull request would turn that into failing way earlier. What do you think?


gbtami (approved):
Agreed.

